PR #15, opened 2024-04-29
Anchor IDL V01

kespinola:
### Goal
Kinobi is able to generate clients for protocols using the latest IDL schema of anchor.

### Changes
- Create v01 transform functions following the updated IDL spec.


kespinola:
@lorisleiva here is the initial scaffolding based on the V00. I'm now going through writing specs and resolving the type violations caused by the updated IDL definition.

packages/nodes-from-anchor/src/index.ts
```diff
@@ -1,12 +1,18 @@
...
 export * from './discriminators';
 export * from './v00';
+export * from './v01';
 
-export type AnchorIdl = IdlV00;
+export type AnchorIdl = IdlV00 | IdlV01;
 
 export function rootNodeFromAnchor(idl: AnchorIdl): RootNode {
-    return rootNodeFromAnchorV00(idl);
+    if ((idl as IdlV01).address) {
```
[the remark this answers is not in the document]
lorisleiva: I think it might be more robust to rely on `idl.metadata.spec` since this is the new field they've introduced to keep track of the version of the spec.

So if `idl?.metadata?.spec` is undefined, we fallback to `v00`, otherwise we can parse that version and create a switch/case for all supported versions. Right now that would only be `"0.1.0"` and we can throw an error for any other version.

kespinola:
@lorisleiva getting somewhere now. List of considerations:

1. New version of IDL does referencing to types from accounts and instructions. I passed types to accounts but lower structs used as args don't have access to the types list.
2. Pdas and seed constant fields.
3. I dropped tuple and map field types. Give these a look could use some guidance.

packages/nodes-from-anchor/src/v01/typeNodes/EnumTypeNode.ts
```diff
@@ -0,0 +1,31 @@
...
+            return enumStructVariantTypeNodeFromAnchorV01(variant);
+        }
+        return enumTupleVariantTypeNodeFromAnchorV01(
+            variant as IdlV01EnumVariant & { fields: IdlV01DefinedFieldsTuple },
+        );
+    });
+    return enumTypeNode(variants);
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
lorisleiva, line 25: nit: looks like we can remove this.

packages/nodes-from-anchor/src/v01/typeNodes/StructFieldTypeNode.ts
```diff
@@ -0,0 +1,26 @@
...
+        });
+    }
+
+    const type = idl as IdlV01Type;
+
+    return structFieldTypeNode({
+        docs: [],
+        name: '',
+        type: typeNodeFromAnchorV01(type),
+    });
```
lorisleiva, line 25: Is this a thing in the new IDL spec? Can field not be named? Or should this be converted into a `TupleTypeNode` instead?

packages/nodes-from-anchor/src/v01/typeNodes/TypeNode.ts
```diff
@@ -0,0 +1,106 @@
...
+        if (idlType === 'string') return sizePrefixTypeNode(stringTypeNode('utf8'), numberTypeNode('u32'));
+        if (idlType === 'bytes') return sizePrefixTypeNode(bytesTypeNode(), numberTypeNode('u32'));
+        return numberTypeNode(idlType);
+    }
+
+    // Ensure eveything else is an object.
+    if (typeof idlType !== 'object') {
+        throw new KinobiError(KINOBI_ERROR__UNRECOGNIZED_ANCHOR_IDL_TYPE, {
+            idlType: JSON.stringify(idlType),
+        });
```
lorisleiva, line 60: nit: maybe we should add a `idlVersion` field on the context of this error. See `packages/errors/src/context.ts`.

packages/nodes-from-anchor/src/v01/typeNodes/TypeNode.ts
```diff
@@ -0,0 +1,106 @@
...
+    // Vec.
+    if ('vec' in idlType) {
+        return arrayTypeNodeFromAnchorV01(idlType);
+    }
+
+    // Defined link.
+    // TODO: Support generics.
+    if ('defined' in idlType && typeof idlType.defined === 'object') {
+        return definedTypeLinkNode(idlType.defined.name);
+    }
```
lorisleiva, line 77: Thanks for adding a TODO for generics here. We can always add that to the Kinobi standard later on.

packages/nodes-from-anchor/test/v01/AccountNode.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    t.deepEqual(
+        node,
+        accountNode({
+            data: structTypeNode([
+                structFieldTypeNode({
+                    name: 'myField',
+                    type: numberTypeNode('u64'),
+                }),
+            ]),
```
lorisleiva, line 38: I'm confused, shouldn't this also have a discriminator field?

packages/nodes-from-anchor/test/v01/typeNodes/OptionTypeNode.test.ts
```diff
@@ -0,0 +1,26 @@
...
+test('it creates option type nodes', t => {
+    t.deepEqual(typeNodeFromAnchorV01({ option: 'u8' }), optionTypeNode(numberTypeNode('u8')));
+});
+
+test('it creates option type nodes with custom prefixes', t => {
+    t.deepEqual(
+        typeNodeFromAnchorV01({ option: 'u8' }),
+        optionTypeNode(numberTypeNode('u8'), { prefix: numberTypeNode('u64') }),
+    );
+});
```
lorisleiva, line 15: I think we can delete that test for v01. I added a `size` to v00 a while ago so it could also support Shank but let's not add it to v01 so it stays true to the Anchor IDL spec.

packages/nodes-from-anchor/test/v01/typeNodes/TupleTypeNode.test.ts
```diff
@@ -0,0 +1,13 @@
+import { numberTypeNode, publicKeyTypeNode, tupleTypeNode } from '@kinobi-so/nodes';
+import test from 'ava';
+
+import { typeNodeFromAnchorV01 } from '../../../src/index.js';
+
+test('it creates tuple type nodes', t => {
+    const node = typeNodeFromAnchorV01({
+        fields: ['u8', 'pubkey'],
+        kind: 'struct',
+    });
```
lorisleiva, line 10: Oh I guess this answers my question how Anchor defines tuples.

packages/nodes-from-anchor/src/discriminators.ts
```diff
@@ -1,6 +1,10 @@
 import { BytesValueNode, bytesValueNode, pascalCase, snakeCase } from '@kinobi-so/nodes';
 import { sha256 } from '@noble/hashes/sha256';
 
+export const getAnchorDiscriminatorV01 = (discriminator: number[]): BytesValueNode => {
+    return bytesValueNode('base16', hex(new Uint8Array(discriminator)));
+};
+
 export const getAnchorInstructionDiscriminator = (idlName: string): BytesValueNode => {
```
lorisleiva, line 8: nit: Maybe we should suffix the other helpers with `V00` as well.

packages/nodes-from-anchor/src/v01/AccountNode.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export function accountNodeFromAnchorV01(idl: IdlV01Account, types: IdlV01TypeDef[]): AccountNode<StructTypeNode> {
+    const definition = types.find(type => type.name === idl.name) as IdlV01TypeDef;
+
+    const idlName = definition.name;
+    const name = camelCase(idlName);
+
+    // TODO: support enum and generic types.
+    let data = structTypeNodeFromAnchorV01(definition.type as IdlV01TypeDefTyStruct);
+    assertIsNode(data, 'structTypeNode');
```
lorisleiva, line 25: Yeah currently I'm not planning on supporting enum type for accounts because we'd loose a lot of the connections within the Kinobi tree if we did that but I'll simmer on it.

lorisleiva (commented):
Thanks a lot for this PR.

> 1. New version of IDL does referencing to types from accounts and instructions. I passed types to accounts but lower structs used as args don't have access to the types list.

I think you did the right thing for the `AccountNode`. As for nested data, Kinobi has a `DefinedTypeLinkNode` that can be used anywhere a `TypeNode` is required.

> 2. Pdas and seed constant fields.

Anchor only defines PDAs in the instruction accounts by re-defining the entire PDA structure. Kinobi offers first-class citizen `PdaNode` attached to `ProgramNode`s which can then be reused inside `InstructionAccountNode` via `PdaLinkNode`s.

I think the best thing to do here is to scan the Anchor PDAs and add `PdaNode` for the current program whenever necessary. If the program ID of the PDA doesn't match the program we are describing, we could add `additionalPrograms` in the `RootNode` such that they would include only `PdaNode`s.

> 3. I dropped tuple and map field types. Give these a look could use some guidance.

I'm not too sure how these are represented in the new Anchor spec. If someone defined a tuple or a map in a Rust struct somewhere, how does Anchor describe them in the IDL?

Note that I've added lot's of things to my To-Do list reading this to make sure the Kinobi IDL doesn't end up having less features than the Anchor IDL.

kespinola:
@lorisleiva thanks for the initial review. I'm looking at handling type references. In the accounts and events lists they are pointers based on the `name` field (Exp 1) and on other fields in instructions the type is `{ "defined": { "name": "StructName" }}` Exp 2. Is there a means of representing these connections in kinobi? I saw the definedlink.

> Accounts and events as type
Account and event type information used to be stored inside their own properties, accounts and events respectively. This created a problem where types weren't being found in some cases because type definitions didn't exist in the types field.

> Now, all defined types are stored in types field, including account an event types. accounts and events field still exist and they act as a pointer to the actual type definition.

## Exp 1
```
{
  "accounts": [{ "name": "MyAccount" }],
  "events": [{ "name": "MyEvent" }],
  "types": [
    {
      "name": "MyAccount"
      // Definition...
    },
    {
      "name": "MyEvent"
      // Definition...
    }
  ]
}
```
## Exp 2
```
        {
          "name": "vec_struct_field",
          "type": {
            "vec": {
              "defined": {
                "name": "FooStruct"
              }
            }
          }
        },
```

packages/nodes-from-anchor/src/v01/typeNodes/StructFieldTypeNode.ts
kespinola:
```typescript
export type IdlV01TypeDefTyStruct = {
    fields?: IdlV01DefinedFields;
    kind: 'struct';
};

export type IdlV01DefinedFields = IdlV01DefinedFieldsNamed | IdlV01DefinedFieldsTuple;

export type IdlV01DefinedFieldsNamed = IdlV01Field[];

export type IdlV01DefinedFieldsTuple = IdlV01Type[];
```

A single field can now be of type field which aligns with the structFieldTypeNode but the IdlV01Type are simple types like 'string', 'u8', etc

kespinola: What do you think on this? I threw an error but the return type should be StructFieldNode or TypeNode.

packages/nodes-from-anchor/src/v01/ProgramNode.ts
```diff
@@ -0,0 +1,26 @@
...
+import { instructionNodeFromAnchorV01 } from './InstructionNode';
+
+export function programNodeFromAnchorV01(idl: IdlV01): ProgramNode {
+    const types = idl.types ?? [];
+    const accounts = idl.accounts ?? [];
+    const instructions = idl.instructions ?? [];
+    const errors = idl.errors ?? [];
+
+    return programNode({
+        accounts: accounts.map(a => accountNodeFromAnchorV01(a, types)),
```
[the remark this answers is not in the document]
kespinola: dropped this but not creating the link to the type just registering the name and discriminator.

kespinola:
All specs passing but still needs work on type linking and pdas.

packages/nodes-from-anchor/src/v01/AccountNode.ts
```diff
@@ -0,0 +1,31 @@
...
+        name: 'discriminator',
+        type: bytesTypeNode(),
+    });
+
+    // TODO: How to set defined type link to the field definition in types list?
+    return accountNode({
+        data: structTypeNode([discriminator]),
+        discriminators: [fieldDiscriminatorNode('discriminator')],
+        name,
+    });
```
kespinola, line 30: In my understanding of defined link type that would create work for a field on the structTypeNode. 

In this case the struct type node can define the discriminator as a field but needs to concat the fields stored on the type 

```suggestion
    // TODO: How to set defined type link to the field definition in types list?
    return accountNode({
        data: structTypeNode([discriminator]),
        discriminators: [fieldDiscriminatorNode('discriminator')],
        references: defnedTypeLink(name)
        name,
    });
```

packages/nodes-from-anchor/src/v01/AccountNode.ts
kespinola: There is a enum type node currently should that be dropped? I need to look over the enum structure some more.

packages/nodes-from-anchor/src/v01/typeNodes/StructFieldTypeNode.ts
```diff
@@ -0,0 +1,23 @@
...
+        return structFieldTypeNode({
+            docs: field.docs ?? [],
+            name: field.name,
+            type: typeNodeFromAnchorV01(field.type),
+        });
+    }
+
+    throw new KinobiError(KINOBI_ERROR__UNRECOGNIZED_ANCHOR_IDL_TYPE, {
+        idlType: JSON.stringify(idl),
+    });
```
kespinola, line 22: Not accounting for the fact that struct objects can have field that are leaf types (so name associated).

```
{
 kind: "struct",
 fields: ["u32", "pubkey"]
}
```

packages/nodes-from-anchor/src/v01/typeNodes/StructTypeNode.ts
```diff
@@ -0,0 +1,10 @@
+import { StructTypeNode, structTypeNode } from '@kinobi-so/nodes';
+
+import { IdlV01DefinedFields, IdlV01TypeDefTyStruct } from '../idl';
+import { structFieldTypeNodeFromAnchorV01 } from './StructFieldTypeNode';
+
+export function structTypeNodeFromAnchorV01(idl: IdlV01TypeDefTyStruct): StructTypeNode {
+    const fields: IdlV01DefinedFields = idl.fields ?? [];
+
+    return structTypeNode(fields.map(structFieldTypeNodeFromAnchorV01));
```
kespinola, line 9: Here the structTypeNode is always expecting field references with a a name and type not just typeFieldNode.

kespinola (commented):
@lorisleiva give this a look. It be helpful to get your review before proceeding further.

kespinola:
I've started setting up the PDAs but current kinibo node catalogue doesn't look to support references other field on an instruction. 

Able to represent the const now but no node types for kind 'account' | 'arg'
```
export type IdlV01Pda = {
    program?: IdlV01Seed;
    seeds: IdlV01Seed[];
};

export type IdlV01Seed = IdlV01SeedAccount | IdlV01SeedArg | IdlV01SeedConst;

export type IdlV01SeedConst = {
    kind: 'const';
    value: number[];
};

export type IdlV01SeedArg = {
    kind: 'arg';
    path: string;
};

export type IdlV01SeedAccount = {
    account?: string;
    kind: 'account';
    path: string;
};
```

packages/nodes-from-anchor/src/v01/typeNodes/StructFieldTypeNode.ts
lorisleiva: Since you're already tackling this use-case in the `typeNodeFromAnchorV01` function, I think it's fine to throw here. I would even only accept the `IdlV01Field` type as argument here and fail higher in the `structTypeNodeFromAnchorV01` helper instead.

I'll comment on the relevant bit of code but if you manage to capture the case were all fields are unnamed for tuples and all fields are named for structs, then we're only throwing if there is a mix and match of named and unnamed fields which I don't think we should aim to support.

EDIT: In fact, based on that IDL type, a mix and match should never occur:

```ts
export type IdlV01DefinedFields = IdlV01DefinedFieldsNamed | IdlV01DefinedFieldsTuple;
export type IdlV01DefinedFieldsNamed = IdlV01Field[];
export type IdlV01DefinedFieldsTuple = IdlV01Type[];
```

packages/nodes-from-anchor/src/v01/typeNodes/TypeNode.ts
```diff
@@ -0,0 +1,113 @@
...
+    if ('coption' in idlType) {
+        return optionTypeNodeFromAnchorV01(idlType as IdlV01TypeCOption);
+    }
+
+    // Struct.
+    if ('kind' in idlType && 'fields' in idlType && idlType.kind === 'struct') {
+        // TODO: bug here where struct type field is getting handled as a tuple if it has 2 fields
+        if (isArrayOfSize(idlType.fields, 2)) {
+            return tupleTypeNodeFromAnchorV01(idlType.fields as IdlV01DefinedFieldsTuple);
+        }
```
lorisleiva, line 100: I think the if-statement catching tuple fields here is perfect but the condition is wrong. `isArrayOfSize` with argument `2` is trying to match on `[x, y]` exactly whereas we need to match on an array of unnamed fields — i.e. `IdlV01Type[]`. So here it would be something like `idlType.fields.every(isIdlFieldV01)`.

packages/nodes-from-anchor/src/v01/AccountNode.ts
lorisleiva: I'll explain this in more details tomorrow morning but this is one of the only place you can't use a `DefinedTypeLinkNode` (for the account data). I may revisit this in the future but `DefinedTypeLinkNode` have the ability to link outside of the tree using the `importFrom` optional attribute. This means we could end up with an `AccountNode` with no data and explorers wouldn't know what do to with it. For now, I think it's best to copy the defined type here and ensure it is a `StructTypeNode` (throwing an error otherwise).

packages/nodes-from-anchor/src/v01/AccountNode.ts
```diff
@@ -0,0 +1,44 @@
...
+    // Account discriminator.
+    const discriminator = structFieldTypeNode({
+        defaultValue: getAnchorDiscriminatorV01(idl.discriminator),
+        defaultValueStrategy: 'omitted',
+        name: 'discriminator',
+        type: bytesTypeNode(),
+    });
+
+    data = structTypeNode([discriminator, ...data.fields]);
+    const discriminators = [fieldDiscriminatorNode('discriminator')];
```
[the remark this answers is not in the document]
lorisleiva: I see this has changed since yesterday but this was perfect. That is: copying over the `StructTypeNode` of the linked `DefinedTypeNode` and prepending the discriminator. 👌

packages/nodes-from-anchor/src/v01/ProgramNode.ts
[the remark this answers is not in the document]
lorisleiva: [Same as this](https://github.com/kinobi-so/kinobi/pull/15#discussion_r1586412516), I think this was the right approach before.

kespinola:
@lorisleiva I think we are in a good spot now. Ready to test on a program and final review.

packages/nodes-from-anchor/src/v01/AccountNode.ts
```diff
@@ -0,0 +1,52 @@
...
+        throw new KinobiError(KINOBI_ERROR__VISITORS__ACCOUNT_FIELD_NOT_FOUND, {});
+    }
+
+    if (type.type.kind !== 'struct') {
+        throw new KinobiError(KINOBI_ERROR__VISITORS__ACCOUNT_FIELD_NOT_FOUND, {});
+    }
+
+    const idlTypeStruct = type.type as IdlV01TypeDefTyStruct;
+
+    const data = structTypeNodeFromAnchorV01(idlTypeStruct);
```
lorisleiva, line 32: I'd probably transform this to:

```ts
const data = typeNodeFromAnchorV01(type.type);
assertIsNode(data, 'structTypeNode');
```

packages/nodes-from-anchor/src/v01/typeNodes/TypeNode.ts
lorisleiva: another nit: Since we're gonna have multiple Anchor errors now, we should create an "Anchor" group of errors starting with `KINOBI_ERROR__ANCHOR__`.

So `KINOBI_ERROR__UNRECOGNIZED_ANCHOR_IDL_TYPE` could be renamed `KINOBI_ERROR__ANCHOR__UNRECOGNIZED_IDL_TYPE`.

packages/nodes-from-anchor/test/v01/typeNodes/TupleTypeNode.test.ts
```diff
@@ -0,0 +1,13 @@
+import { numberTypeNode, publicKeyTypeNode, tupleTypeNode } from '@kinobi-so/nodes';
+import test from 'ava';
+
+import { typeNodeFromAnchorV01 } from '../../../src/index.js';
+
+test('it creates tuple type nodes', t => {
```
lorisleiva: We should probably add a test here that makes sure we throw an error when we end up with a Anchor IDL struct with a mixed of named and unnamed fields.

packages/nodes-from-anchor/src/v01/ProgramNode.ts
```diff
@@ -0,0 +1,34 @@
...
+    const types = idl.types ?? [];
+    const accounts = idl.accounts ?? [];
+    const instructions = idl.instructions ?? [];
+    const errors = idl.errors ?? [];
+
+    const definedTypes = types.map(definedTypeNodeFromAnchorV01);
+    const accountNodeFromAnchorV01 = accountNodeFromAnchorV01WithTypeDefinition(types);
+    const pdas = instructions
+        .flatMap<IdlV01InstructionAccount>(instruction => instruction.accounts)
+        .filter(account => !account.pda?.program)
```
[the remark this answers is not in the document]
kespinola: Thanks. Added case to cover the previous failure.

kespinola:
@lorisleiva CR is in.

lorisleiva:
Thanks! There are a few final touches to fix but I can sort this out in a subsequent commit. 🍻

PR #27, opened 2024-05-10
make rust docs kinobi link clickable

samuelvanderwaal:
Makes the Kinobi link clickable in the Rust docs and avoids the following compilation warning:

Screenshot 2024-05-10 at 8 37 10 AM


lorisleiva (approved):
Nice, I wasn't aware of this. Thank you! 🍺

PR #32, opened 2024-05-16
Add default impl for instruction data struct

febo:
This PR adds a `Default` impl for instruction data structs.

### DIFF
--- a/packages/renderers-rust/test/instructionsPage.test.ts
+++ b/packages/renderers-rust/test/instructionsPage.test.ts
@@ -47,3 +47,18 @@ test('it renders an instruction with a remainder str', () => {
         `pub memo: RemainderStr`,
     ]);
 });
+
+test('it renders a default impl for instruction data struct', () => {
+    // Given the following program with 1 instruction.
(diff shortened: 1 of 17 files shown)


packages/renderers-rust/test/instructionsPage.test.ts
```diff
@@ -47,3 +47,21 @@ test('it renders an instruction with a remainder str', () => {
...
+    const node = programNode({
+        instructions: [instructionNode({ name: 'mintTokens' })],
+        name: 'splToken',
+        publicKey: 'TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA',
+    });
+
+    // When we render it.
+    const renderMap = visit(node, getRenderMapVisitor());
+
+    // Then we expect the following pub struct.
```
lorisleiva, line 62:
```suggestion
    // Then we expect the Default trait to be implemented.
```

lorisleiva (approved):
Thanks!

lorisleiva (approved):

PR #37, opened 2024-05-17
Remove Anchor V01 Accounts from Defined Types

kespinola:
resolves https://github.com/kinobi-so/kinobi/issues/36

## Changes
- Remove types from the anchor idl that match up with accounts before generating definedTypes.

### DIFF
--- a/.changeset/hip-years-clean.md
+++ b/.changeset/hip-years-clean.md
@@ -1,0 +1,5 @@
+---
+"@kinobi-so/nodes-from-anchor": patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/hip-years-clean.md
```diff
@@ -0,0 +1,6 @@
+---
+"@kinobi-so/nodes-from-anchor": patch
+"@kinobi-so/nodes": patch
```
lorisleiva, line 3: I think we can remove this line so the version bump only affects `nodes-from-anchor`.

lorisleiva (approved):
Thanks!

lorisleiva (approved):

PR #42, opened 2024-05-21
Fully Support Anchor 0.30.0 IDL Seed Spec

kespinola:
## Background
The original release of anchor 0.30.0 only shipped with basic account and const support.

## Goal
Support the complete seed specification.

## Changes
- PDA extraction is based on instructions and complete list of types.
- PDA nodes are scoped by the instruction `<instruction><account_name>`
- Add support for struct field reference found in top level types.
- Set seed variable based on instruction argument.

## Next Step
- Incorporate the PDA into the instruction so addresses computed automatically based on the lookup up schema set in the IDL.


lorisleiva (commented):
Thank you for that! I've left some comments, let me know what you think.

packages/nodes-from-anchor/src/v01/PdaNode.ts
```diff
@@ -1,39 +1,102 @@
...
+    IdlV01TypeDef,
+    IdlV01TypeDefTyStruct,
+} from './idl';
+import { typeNodeFromAnchorV01 } from './typeNodes';
+
+export function pdaNodesFromAnchorInstructionV01(instruction: IdlV01Instruction, types: IdlV01TypeDef[]): PdaNode[] {
+    return instruction.accounts.reduce((memo: PdaNode[], account: IdlV01InstructionAccount) => {
+        const pda = account.pda;
+        if (pda) {
+            const name = camelCase(`${instruction.name}_${account.name}`);
```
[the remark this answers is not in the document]
kespinola: > Also note that we could update the PdaValueNode such that its pda attribute could either be a PdaLinkNode or a PdaNode directly. That way each instruction account would be responsible for defining its own PdaNode. Then, we can use the defaultVisitor to goes through each of these inlined PdaNode and transform them into PdaLinkNode when applicable.

I'm up to explore this approach. At least for how anchor defines pda they shouldn't be directly exposed to the end caller. The IDL allows for the seeds to be computed off of the context of the instruction. I'm actually thinking that they shouldn't be included in Kinobi program pda list as they are helps generally defined on account independently of the instruction.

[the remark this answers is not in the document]
lorisleiva: It makes sense, I don't think it would be that big of a change and it wouldn't even be a breaking change. I can have a look at it tomorrow by working on your branch. I'll also need to add an optional `programId` attribute to the `PdaNode` so it can derive from other programs too.

I still think it could be valuable to have a visitor that goes through the whole tree afterwards and try to extract some of these nested `PdaNodes` into top-level `PdaNodes` so the end-user has helpers to create these PDAs. We would just need to be extra-strict on the ones we extract. For instance, we would only extract the ones that are from this program and we would make sure we deduplicate any PDA that's used in multiple instructions. But the beauty of having that on its own visitor is people can opt-out of this whole logic by opting out of the default visitors.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
```diff
@@ -22,6 +36,40 @@ export function instructionAccountNodeFromAnchorV01(idl: IdlV01InstructionAccoun
...
+                        ];
+                    case 'account':
+                        return [
+                            [...seeds, variablePdaSeedNode(seed.path, publicKeyTypeNode())],
+                            [...lookups, pdaSeedValueNode(seed.path, accountValueNode(seed.path))],
+                        ];
+                    default:
+                        throw new KinobiError(KINOBI_ERROR__ANCHOR__SEED_KIND_UNIMPLEMENTED, {
+                            kind,
+                        });
```
kespinola, line 58: At this point, argument lookup is not supported because the account helper doesn’t have access to the arguments (though it can be updated to do so).

@lorisleiva, is Kinobi good to allow the type node for the argument field using variablePdaSeedNode(seed.path, typeNode(arg.type))?

Additionally, we are not handling the injection of seeds from deserialized accounts as we discussed, so treasury.mint support is not included, for example.

Also not handling injecting seed from deserialized account as we discussed so no `treasury.mint` support for example.

kespinola (commented):
I tried out an e2e test by generating clients for wpl programs but node only wants pda link which you changes so may not be referencing the packages correctly. I did fresh build.

```
// package.json
"dependencies": {
    "@kinobi-so/nodes": "link:../kinobi/packages/nodes",
    "@kinobi-so/nodes-from-anchor": "link:../kinobi/packages/nodes-from-anchor",
    "@kinobi-so/renderers": "link:../kinobi/packages/renderers",
    "@kinobi-so/renderers-core": "link:../kinobi/packages/renderers-core",
    "@kinobi-so/visitors": "link:../kinobi/packages/visitors",
    "@kinobi-so/visitors-core": "link:../kinobi/packages/visitors-core",}
```


```javascript
// kinobi.mjs
import path from "path";
import { fileURLToPath } from "url";
import {
  renderRustVisitor,
  renderJavaScriptVisitor,
} from "@kinobi-so/renderers";
import { rootNodeFromAnchor } from "@kinobi-so/nodes-from-anchor";
import { readJson } from "@kinobi-so/renderers-core";
import { visit } from "@kinobi-so/visitors-core";
import fs from "fs/promises";

const __dirname = path.dirname(fileURLToPath(import.meta.url));
const clientDir = path.join(__dirname, "clients");
const idlDir = path.join(__dirname, "target", "idl");

const idlFiles = await fs.readdir(idlDir);

for (const idlFile of idlFiles) {
  const idlPath = path.join(idlDir, idlFile);
  const idl = readJson(idlPath);

  const node = rootNodeFromAnchor(idl);

  const sdkName = idl.metadata.name;

  await visit(
    node,
    renderJavaScriptVisitor(
      path.join(clientDir, "js", sdkName, "src", "generated")
    )
  );

  visit(
    node,
    renderRustVisitor(
      path.join(clientDir, "rust", sdkName, "src", "generated"),
      { format: true }
    )
  );
}

```

```
    throw new KinobiError(KINOBI_ERROR__UNEXPECTED_NODE_KIND, {
          ^

KinobiError: Kinobi error #2; Decode this error by running `npx @kinobi-so/errors decode -- 2 'ZXhwZWN0ZWRLaW5kcz0lNUJwZGFMaW5rTm9kZSU1RCZraW5kPXBkYU5vZGUmbm9kZT0lNUJvYmplY3QlMjBPYmplY3QlNUQ='`
    at assertIsNode (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/nodes/dist/index.node.mjs:586:11)
    at Object.visitPdaValue (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:596:7)
    at visit (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:152:22)
    at file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:178:72
    at Object.visitInstructionAccount (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:239:60)
    at visit (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:152:22)
    at file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:178:72
    at Array.map (<anonymous>)
    at Object.visitInstruction (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:227:33)
    at visit (file:///Users/kyleespinola/Code/abk-labs/kinobi/packages/visitors-core/dist/index.node.mjs:152:22) {
  context: {
    __code: 2,
    expectedKinds: [ 'pdaLinkNode' ],
    kind: 'pdaNode',
    node: {
      kind: 'pdaNode',
      name: 'group',
      docs: [],
      seeds: [
        {
          kind: 'constantPdaSeedNode',
          type: { kind: 'bytesTypeNode' },
          value: {
            kind: 'bytesValueNode',
            data: '67726f7570',
            encoding: 'base16'
          }
        },
        {
          kind: 'variablePdaSeedNode',
          name: 'groupMint',
          docs: [],
          type: { kind: 'publicKeyTypeNode' }
        }
      ]
    }
  }
}

Node.js v19.8.1
 ELIFECYCLE  Command failed with exit code 1.
➜  wen-program-library git:(main) ✗ npx @kinobi-so/errors decode -- 2 'ZXhwZWN0ZWRLaW5kcz0lNUJwZGFMaW5rTm9kZSU1RCZraW5kPXBkYU5vZGUmbm9kZT0lNUJvYmplY3QlMjBPYmplY3QlNUQ='

[Decoded] Kinobi error code #2
    - Expected node of kind [[pdaLinkNode]], got [pdaNode].

[Context]
    {
        "expectedKinds": "[pdaLinkNode]",
        "kind": "pdaNode",
        "node": "[object Object]"
    }
```

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
lorisleiva: Regarding the argument, yes you should be able to do something like this:

```ts
case 'argument':
    return [
        [...seeds, variablePdaSeedNode(seed.path, typeNodeFromAnchorV01(arg.type))],
        [...lookups, pdaSeedValueNode(seed.path, argumentValueNode(seed.path))],
    ];
```

You might just want to have an if statement first making sure `!seed.path.includes('.')`.

packages/errors/src/context.ts
```diff
@@ -50,8 +54,21 @@ type DefaultUnspecifiedErrorContextToUndefined<T> = {
...
+    };
+    [KINOBI_ERROR__ANCHOR__SEED_KIND_UNIMPLEMENTED]: {
+        kind: string;
+    };
+    [KINOBI_ERROR__ANCHOR__TYPE_PATH_MISSING]: {
         idlType: string;
+        path: string;
+    };
+    [KINOBI_ERROR__ANCHOR__UNRECOGNIZED_IDL_TYPE]: {
+        name: CamelCaseString;
```
lorisleiva, line 71: The context of this error type shouldn't be changed as it's already being used in the code.

```suggestion
        idlType: string;
```

lorisleiva:
Regarding the e2e tests, you shouldn't need to update dependencies or use relative path since the monorepo will understand they are local packages. You just need to make sure you are running the e2e tests from the package directory — e.g. `cd packages/renderers-js && pnpm test:e2e`.

One thing you could do is duplicate the `system` e2e folder for example and name it `system-from-anchor`. Then you just need to update the `generateProject` function so it uses `rootNodeFromAnchor` instead of `rootNode`.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
kespinola: Ok. Account for args then.

And for path based can it be kept as a variable but with nod pdaSeedValue? Kinobi keeps track of it has the correct type but caller is responsible for passing value?

kespinola:
> Regarding the e2e tests, you shouldn't need to update dependencies or use relative path since the monorepo will understand they are local packages. You just need to make sure you are running the e2e tests from the package directory — e.g. `cd packages/renderers-js && pnpm test:e2e`.
> 
> One thing you could do is duplicate the `system` e2e folder for example and name it `system-from-anchor`. Then you just need to update the `generateProject` function so it uses `rootNodeFromAnchor` instead of `rootNode`.

I'll get a e2e test in place. I started on it before but stopped.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
lorisleiva: Interesting... You would need to add a new entry to the `extraArguments` array of the `instructionNode`. Then you can use this as an `argumentValueNode` in your PDA default value. That `extraArguments` array basically allows you to require more data to the end user when they create their instruction without including them in the instruction data.

That being said, for complex PDAs, it could end up adding lot's of attribute to the instruction input and make things worst so whilst we can do it, I'm not 100% sure that we should. Your call though. 😊

kespinola: Ok @lorisleiva though I could be wrong but I don't thinks its strategy used often but do think having the caller pass as the argument is the best way to handle this case. 

Also for args what are your thoughts and adding a link type field for argument types? 
 `typeArgumentLinkNode(seed.path)` instead of requiring the args to be passed to the `instructionAccountNodeFromAnchorV01` func.

kespinola:
> > Regarding the e2e tests, you shouldn't need to update dependencies or use relative path since the monorepo will understand they are local packages. You just need to make sure you are running the e2e tests from the package directory — e.g. `cd packages/renderers-js && pnpm test:e2e`.
> > One thing you could do is duplicate the `system` e2e folder for example and name it `system-from-anchor`. Then you just need to update the `generateProject` function so it uses `rootNodeFromAnchor` instead of `rootNode`.
> 
> I'll get a e2e test in place. I started on it before but stopped.

And FYI on this. I was linking to external project.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
lorisleiva: Hey, sorry I forgot to answer this one.

Currently all link nodes point to top-level nodes such as programs, instructions, etc. I think it may be confusing to introduce this sort of low-level link node. There may be a better way to avoid that redundancy though like introducing a different PDA value node altogether. I think we can revisit this in a subsequent PR though.

kespinola:
Sorry on lagging on this @lorisleiva. It now handles all seed permutations.

kespinola:
@lorisleiva thanks for fixing up the lock. I am good to merge when you are.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
```diff
@@ -1,25 +1,121 @@
...
+                            const fieldNode = accountFields.find(({ name }) => name === fieldName);
+                            if (!fieldNode) {
+                                throw new KinobiError(KINOBI_ERROR__ANCHOR__TYPE_PATH_MISSING, {
+                                    idlType: seed.account,
+                                    path: seed.path,
+                                });
+                            }
+
+                            const seedName = camelCase(seed.path);
+                            return [[...seeds, variablePdaSeedNode(seedName, fieldNode.type)], []];
```
lorisleiva, line 92: @kespinola I am a bit concerned about this use-case because we are identifying a variable seed for the PDA but we're not able to provide a value for it since the nested account isn't in the instruction accounts.

I initially tried to use `path[0]` as an `accountValueNode` here but then realised that this is not the account we should use, this is just the account that should contain the seed inside of its data which we do not have access to.

So in practice, generating a client for this use-case will throw an error because we'll create a PDA without all the required seeds. I'm not too sure if to leave this scenario as-is and revisit in the future or if to stop supporting this scenario altogether to avoid people getting confused. Wdyt?

kespinola: In this case my thought was to map thise to a variable of the instruction to be provided. It is a very common pattern in anchor IDLs so even if doesn't have auto-lookup it should be skipped gracefully.

I am also fine with dropping any generation of seeds since they aren't mapped to a top level account definition like in the shank IDL spec.

lorisleiva: Ah I see yeah we could use an `extraArgument` for this purpose. For now though how would we gracefully skip this scenario? Should we just discard the whole seed default value if we encounter a seed with a path length greater than 1?

kespinola: Yes. that seems appropriate.

lorisleiva: I've updated the PR accordingly. Let me know if that works for you.

lorisleiva (approved):
LGTM! I made a few tweaks today so let me know if you're all happy with them.

PR #45, opened 2024-05-30
Add copy derive to scalar enums

febo:
This PR adds a `Copy` derive to scalar enums when rendering Rust clients.

### DIFF
--- a/packages/renderers-rust/test/_setup.ts
+++ b/packages/renderers-rust/test/_setup.ts
@@ -10,3 +10,14 @@ export function codeContains(actual: string, expected: (RegExp | string)[] | Reg
         }
     });
 }
+
+export function codeNotContains(actual: string, expected: (RegExp | string)[] | RegExp | string) {
+    const expectedArray = Array.isArray(expected) ? expected : [expected];
+    expectedArray.forEach(e => {
+        if (typeof e === 'string') {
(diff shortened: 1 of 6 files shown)


packages/renderers-rust/test/_setup.ts
```diff
@@ -10,3 +10,14 @@ export function codeContains(actual: string, expected: (RegExp | string)[] | Reg
         }
     });
 }
+
+export function codeNotContains(actual: string, expected: (RegExp | string)[] | RegExp | string) {
```
lorisleiva, line 14: nit: I called this `codeDoesNotContain` on the JS side IIRC. Should we keep the same name to stay consistent?

lorisleiva (approved):
Thanks! Just a small nit but looks good!

kespinola (approved):

lorisleiva:
Thanks! Btw, I'm confused as to why @kespinola's approval is grey since he's a member of the Kinobi org. @febo Do we need to give him special permissions somewhere so that his approval trigger the rule branch protections?

febo:
> Thanks! Btw, I'm confused as to why @kespinola's approval is grey since he's a member of the Kinobi org. @febo Do we need to give him special permissions somewhere so that his approval trigger the rule branch protections?

The base permission for members was set to `read` – I fixed it now.

PR #49, opened 2024-06-05
Update programErrors.njk

wjthieme:
In certain conditions, `__DEV__` is not defined on the global object (for example when not using a bundler). This can cause a fatal error to occur unless `__DEV__` is defined. This for example happens when trying to import the error file from mocha or jest (without the right env)

```
ReferenceError: __DEV__ is not defined
```

This PR merely fixes this fatal error that can be experienced downstream but does not offer any improvements/alternatives. In the future might be worth to make this more inclusive but there does not seem to really be a standard:
* `process.env.NODE_ENV` in nodejs
* `__DEV__` for certain bundlers
* `process.env.VERCEL_ENV` in vercel
* etc.

### DIFF
--- a/packages/renderers-js/public/templates/fragments/programErrors.njk
+++ b/packages/renderers-js/public/templates/fragments/programErrors.njk
@@ -11,7 +11,7 @@ export type {{ programErrorUnion }} =
   {% endfor %};
 
 let {{ programErrorMessagesMap }}: Record<{{ programErrorUnion }}, string> | undefined;
-if (__DEV__) {
+if ("__DEV__" in global && __DEV__) {
   {{ programErrorMessagesMap }} = {
     {% for error in errors | sort(false, false, 'name') %}
       [{{ getProgramErrorConstant(error.name) }}]: `{{ error.message }}`,
@@ -20,7 +20,7 @@ if (__DEV__) {
 }
 
 export function {{ programGetErrorMessageFunction }}(code: {{ programErrorUnion }}): string {
-  if (__DEV__) {
+  if ("__DEV__" in global && __DEV__) {
     return ({{ programErrorMessagesMap }} as Record<{{ programErrorUnion }}, string>)[code];
   }
 


lorisleiva (commented):
Hi there,

This is because, using this renderer, you need to add a compilation step that builds the JavaScript client with this variable.

See https://github.com/solana-program/system/blob/main/clients/js/env-shim.ts

wjthieme:
Thanks for the response! The way I am building the js client now is:
* Generate ts code using kinobi
* Transpile to js using tsc (which mirrors the `__DEV__` check into the js code)
* Publish the transpiled js only to npm which can then be consumed downstream

In this case if `__DEV__` is not defined downstream this error occurs (unless some sort of bundler is used that defines `__DEV__`. I guess if I swapped out the tsc step for some sort of bundler I wouldn't have this problem but I'd rather just use tsc (unless there is a valid reason to not do so of course)

wjthieme:
For anyone with the same problem. I've added an index file that wraps the generated code by kinobi for building with tsc. Not ideal but it works:

```
Object.assign(global, { __DEV__: process.env.NODE_ENV === "development" });

export * from "./generated";
```

lorisleiva:
Thanks for providing a workaround in case other are facing the same issue.

I think we may revisit in the future since it'll make bundling generated packages easier but I'm worried the proposed solution will override the `__DEV__` variable offered by some bundlers such as Webpack.

My recommended way to approach this is to use the [`create-solana-program` template generator](https://github.com/solana-program/create-solana-program) which already offers a way to bundle the generated code and includes a shim for the `__DEV__` variable.

lorisleiva:
Hey 👋 

Sorry I forgot to report back a while ago but this has been fixed by https://github.com/kinobi-so/kinobi/pull/103.

The fix is basically to ditch the `__DEV__` global variable altogether and generate the condition `process.env.NODE_ENV !== "production"` directly.

I hope this helps. 🙂

PR #50, opened 2024-06-06
Add anchor traits impl for accounts

febo:
This PR adds Anchor traits impl to account types so they can be directly referenced in structs annotated with `#[account]` in Anchor programs.

This provides basic validation that:
1. the account must be initialized; and
2. it has the correct owner.


kespinola (approved):
Thank you!

PR #52, opened 2024-06-11
fix: render PDA helper with programAddress = node.programId if defined 

lithdew:
PDA helpers should generate PDA's with a default programAddress set to the provided PDA nodes' programId if programId is defined. If not defined, the default programAddress should be the encapsulating program node programId.

With this PR, the following test passes. All tests are passing locally. A changeset has been included.

```ts
test('it renders a PDA helper function with a default program address', async () => {
    // Given the following PDA node with a default program address.
    const node = programNode({
        name: 'myProgram',
        pdas: [
            pdaNode({
                name: 'foo',
                programId: 'myProgramId',
                seeds: [constantPdaSeedNodeFromString('utf8', 'myPrefix')],
            }),
        ],
        publicKey: '1111',
    });

    // When we render it.
    const renderMap = visit(node, getRenderMapVisitor());

    // Then we expect the following PDA function using the programId as the default program address.
    await renderMapContains(renderMap, 'pdas/foo.ts', [
        'export async function findFooPda',
        "const { programAddress = 'myProgramId' as Address<'myProgramId'> } = config;",
    ]);
});
```

### DIFF
--- a/.changeset/polite-cats-wash.md
+++ b/.changeset/polite-cats-wash.md
@@ -1,0 +1,5 @@
+---
+'@kinobi-so/renderers-js': patch
+---
+
+Fix rendering PDA helper functions with a default programId.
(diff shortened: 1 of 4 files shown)


lorisleiva (approved):
Perfect PR, tysm! 🍺

PR #53, opened 2024-06-14
Filter out type variant on imports

febo:
This PR modifies the `ImportMap` to filter out `type Type` variants from imports when the import for `Type` is present.

Instead of rendering with a duplicated import of `Key`:
```typescript
import { Key, getKeyDecoder, getKeyEncoder, type Key } from '../types';
```
The renderer will output:
```typescript
import { Key, getKeyDecoder, getKeyEncoder } from '../types';
```

### DIFF
--- a/.changeset/shy-points-march.md
+++ b/.changeset/shy-points-march.md
@@ -2,4 +2,4 @@
 '@kinobi-so/renderers-js': patch
 ---
 
-Add missing type prefix for enums
+Filter out type variant on imports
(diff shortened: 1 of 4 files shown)


lorisleiva:
Hmm in this case I think we need the enum but not as a type. Because we're renderer `MyEnum.myVariant` so we need more than the type.

I think we should add a post-processing step on the ImportMap when outputting it as a string such that if `X` and `type X` are imported, it only imports `X`.

PR #109, opened 2024-07-15
Fix Arg Lookup for Anchor V01 Seed Default Value Generation

kespinola:
## Changes
- Convert name to camel case before searching argument node list as kinobi tracks all names in camel case format.


lorisleiva (approved):
Nice catch, thank you!

PR #127, opened 2024-07-21
Add isProgramError helper in JS renderer

lorisleiva:
This PR generates a new helper function in the JS renderer that wraps `isProgramError` from `@solana/programs` such that, the error code type is known and the program address must not be provided.

Here's an example using the system program.

```ts
// Before.
isProgramError(error, txMessage, SYSTEM_PROGRAM_ADDRESS, SYSTEM_ERROR__ACCOUNT_ALREADY_IN_USE);

// After.
isSystemError(error, txMessage, SYSTEM_ERROR__ACCOUNT_ALREADY_IN_USE);
```

Here's another example without passing an explicit code expectation. For it to work with `isProgramError` a type parameter must be provided. With the generated helper, this is no longer needed.

```ts
// Before.
isProgramError<SystemError>(error, txMessage, SYSTEM_PROGRAM_ADDRESS);

// After.
isSystemError(error, txMessage);
```

### DIFF
--- a/packages/renderers-js/e2e/system/test/transferSol.test.ts
+++ b/packages/renderers-js/e2e/system/test/transferSol.test.ts
@@ -3,13 +3,18 @@ import {
   type Address,
   appendTransactionMessageInstruction,
   generateKeyPairSigner,
+  isSolanaError,
   lamports,
   pipe,
+  SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE,
+  SolanaError,
 } from '@solana/web3.js';
 import test from 'ava';
 import {
   getTransferSolInstruction,
+  isSystemError,
   parseTransferSolInstruction,
+  SYSTEM_ERROR__RESULT_WITH_NEGATIVE_LAMPORTS,
 } from '../src/index.js';
 import {
   createDefaultSolanaClient,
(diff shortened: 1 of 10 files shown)


packages/renderers-js/e2e/system/test/transferSol.test.ts
```diff
@@ -46,6 +51,54 @@ test('it transfers SOL from one account to another', async (t) => {
...
+  // Then we expect the transaction to fail and to have the correct error code.
+  const error = await t.throwsAsync(promise);
+  if (
+    isSolanaError(
+      error,
+      SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE
+    )
+  ) {
+    if (isSystemError(error.cause, txMessage)) {
+      error satisfies SolanaError;
```
steveluscher, line 80: Was this meant to be something like this?

```suggestion
    error satisfies SolanaError<typeof SOLANA_ERROR__JSON_RPC__SERVER_ERROR_SEND_TRANSACTION_PREFLIGHT_FAILURE>;
    if (isSystemError(error.cause, txMessage)) {
        error.cause satisfies SolanaError<typeof SOLANA_ERROR__INSTRUCTION_ERROR__CUSTOM> & {
            context: { code: SystemError };
        };
```

packages/renderers-js/public/templates/fragments/programErrors.njk
```diff
@@ -26,3 +26,11 @@ export function {{ programGetErrorMessageFunction }}(code: {{ programErrorUnion
 
   return 'Error message not available in production bundles.';
 }
+
+export function {{ programIsErrorFunction }}<TProgramErrorCode extends {{ programErrorUnion }}>(
+    error: unknown,
+    transactionMessage: { instructions: Record<number, { programAddress: Address }> },
+    code?: TProgramErrorCode,
+): error is SolanaError<typeof SOLANA_ERROR__INSTRUCTION_ERROR__CUSTOM> & { context: { code: TProgramErrorCode } } {
```
steveluscher, line 34: Maybe a dollop of `Readonly<>` on that return type?
```suggestion
): error is SolanaError<typeof SOLANA_ERROR__INSTRUCTION_ERROR__CUSTOM> & Readonly<{ context: Readonly<{ code: TProgramErrorCode }> }> {
```

steveluscher (approved):
identify-jean-luc-picard

packages/renderers-js/public/templates/fragments/programErrors.njk
lorisleiva: Good shout, we should probably do that in `@solana/programs` as well because that was copy/pasted from there.

packages/renderers-js/e2e/system/test/transferSol.test.ts
lorisleiva: Yeah sorry my type checks were not quite as thorough haha. I've updated them. ✅

PR #138, opened 2024-07-24
Use ReadonlyUint8Array for identifying accounts/instructions by data

mcintyre94:
The `identifyXInstruction` and `identifyXAccount` already accept a `Uint8Array` that can be used to identify an instruction given just the data byte array. But if this comes from a web3js encoder (eg encoding base64 data from somewhere else), then the datatype will be `ReadonlyUint8Array` instead. 

This PR updates these identify functions to allow an input of `ReadonlyUint8Array` instead. This doesn't require any further changes, as everything being done with data is already compatible with `ReadonlyUint8Array`. It's also backward compatible, you can still pass a `Uint8Array` in. 

This means that output from our encoders can now be passed directly to these generated instructions.

### DIFF
--- a/.changeset/shaggy-shrimps-call.md
+++ b/.changeset/shaggy-shrimps-call.md
@@ -1,0 +1,5 @@
+---
+'@kinobi-so/renderers-js': patch
+---
+
+Use ReadonlyUint8Array for identifying accounts/instructions by data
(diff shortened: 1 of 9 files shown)


lorisleiva (approved):
Perfect, thanks for this! 🍺

PR #144, opened 2024-07-31
Add support for ShortU16 and ShortVec on Rust renderer

lorisleiva:


packages/renderers-rust/src/getTypeManifestVisitor.ts
```diff
@@ -92,6 +92,13 @@ export function getTypeManifestVisitor(options: { nestedStruct?: boolean; parent
                                     type: `${prefixFormat}PrefixVec<${childManifest.type}>`,
                                 };
                             }
+                            case 'shortU16': {
+                                childManifest.imports.add('solana_program::short_vec::ShortVec');
```
buffalojoec, line 96: I think this is good for now, but flagging that it's going to be deprecated in 2.1.0 in favor of the new crate `solana-short-vec`.

buffalojoec (approved):

packages/renderers-rust/src/getTypeManifestVisitor.ts
lorisleiva: FML 😅

PR #152, opened 2024-08-01
Support type aliases from DefinedTypeNodes in Rust renderer

lorisleiva:
Currently, the Rust renderer can only render `DefinedTypeNodes` if the child type is a `StructTypeNode` or an `EnumTypeNode`.

This PR allows any other child type to be rendered as a Rust type alias like so.

```rust
pub type MyAlias = ShortVec<Pubkey>;
```


febo (approved):
Perfect!

PR #161, opened 2024-08-07
Render sol amount as lamports

mcintyre94:
Note:  #160 is a prerequisite (Graphite 💔)

This PR updates how `SolAmountTypeNode` is rendered in the new web3js renderer. Previously it just rendered its nested number type. It now renders with a `Lamports` type from web3js. The encoder + decoder use `getLamportsEncoder` and `getLamportsDecoder`, passing the nested number encoder/decoder as an argument.

This means that you can represent Lamports in the Kinobi IDL using any numeric type, and you'll get properly typed Lamports back. This can improve eg parsing of instructions like `createAccount` where some arguments represent Lamport values.

### DIFF
--- a/.changeset/cyan-badgers-stare.md
+++ b/.changeset/cyan-badgers-stare.md
@@ -1,0 +1,5 @@
+---
+'@kinobi-so/renderers-js': patch
+---
+
+Use lamports type/encoder/decoder for SolAmountTypeNode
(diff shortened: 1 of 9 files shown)


lorisleiva:
Nice thanks for this! Would you mind adding a changeset and a test for this?

For the changeset, you can just run `changeset` locally and for the test you can [get inspired by this test](https://github.com/kinobi-so/kinobi/blob/main/packages/renderers-js/test/types/sizePrefix.test.ts) and name it something like `solAmount.test.ts`.

lorisleiva (approved):
Perfect, thank you!

PR #163, opened 2024-08-09
Expose struct default/optional values as constants in renderers-js

wjthieme:
This is particularly useful for when you require an account or instructionData discriminator downstream. An example would be when writing GPA filters.

@lorisleiva ~~I still need to figure out how to add the struct name to the constant. Right now if the field that has a default value is called `discriminator` the constant will be called `DISCRIMINATOR`. This is problematic since just `DISCRIMINATOR` is not very descriptive and it will conflict with other constants. I'd rather have it something like `${structName}_`${fieldName}` but from `visitStructType` I do not have the name of the struct itself (or I missed it somewhere). Any idea how I would be able to add this?~~ Any other feedback is always welcome too!

This change alters the rendering of the `type` fragment. An example below:

before
```ts
export type AllocateInstructionData = { discriminator: number; space: bigint };

export type AllocateInstructionDataArgs = { space: number | bigint };

export function getAllocateInstructionDataEncoder(): Encoder<AllocateInstructionDataArgs> {
  return transformEncoder(
    getStructEncoder([
      ['discriminator', getU32Encoder()],
      ['space', getU64Encoder()],
    ]),
    (value) => ({ ...value, discriminator: 8 })
  );
}
```

After
```ts
export const ALLOCATE_INSTRUCTION_DATA_DISCRIMINATOR = 8;

export type AllocateInstructionData = { discriminator: number; space: bigint };

export type AllocateInstructionDataArgs = { space: number | bigint };

export function getAllocateInstructionDataEncoder(): Encoder<AllocateInstructionDataArgs> {
  return transformEncoder(
    getStructEncoder([
      ['discriminator', getU32Encoder()],
      ['space', getU64Encoder()],
    ]),
    (value) => ({ ...value, discriminator: ALLOCATE_INSTRUCTION_DATA_DISCRIMINATOR })
  );
}
```


lorisleiva:
Hi there, Thanks for your contribution. I like the overall idea but my issue with this approach is that every single default value will end up extracted as a constant which may result in lots of non-discriminator constants polluting the library's namespace.

I wonder if should instead use the discriminator nodes to explicitly extract these constant. This may not be trivial though as there is a `bytesDiscriminatorNode` that doesn't link the default value of a data field.

wjthieme:
> I wonder if should instead use the discriminator nodes to explicitly extract these constant. This may not be trivial though as there is a bytesDiscriminatorNode that doesn't link the default value of a data field.

Yeah initially I wanted to figure out a solution to just export the Discriminator but couldn't find an easy way to do this. I'd say the positives (having the Discriminator exported) outweigh the negatives (exporting all default values which IMO is not necessarily a downside). Having the discriminator of accounts/instructions is pretty vital when doing gpa or quickly filtering through a whole bunch of instructions. 

Currently what I would need to do to get the discriminator is to encode the full struct and then extract the constant bytes.

lorisleiva:
I hear what you are saying but I disagree that it outweighs the negatives. There are programs using Kinobi with lots of defaults values such as `none()` or `0` and I can imagine these programs maintainers not being super pleased about having dozens of useless constant exports (like `MY_IX_OPTION_A = none()`) for each of their instruction.

Once something like this is merged, it is very difficult to revert as it'll be a breaking change. I say, let's do things right the first time around, even if that means it takes a bit longer to ship.

wjthieme:
> Once something like this is merged, it is very difficult to revert as it'll be a breaking change. I say, let's do things right the first time around, even if that means it takes a bit longer to ship.

A quick solution would be to only export the default values where `defaultValueStrategy === 'omitted'`. Those are the only ones anyone would care about probably

lorisleiva:
This still feels like the wrong approach as we're relying on any default values to export discriminators. How about we tackle this using discriminator nodes instead?

The process would be as follow:
1. Create a new fragment that specialises in exporting discriminator constants.
2. This new fragment goes through all `DiscriminatorNode`s of the instruction such that:
    - If it is a `fieldDiscriminatorNode`, it exposes a variable of name `IX_NAME_FIELD_NAME` and uses the field's default value as the constant value (perhaps as a `Uint8Array` for consistency?).
    - If it is a `bytesDiscriminatorNode`, it exposes a variable of name `IX_NAME_DISCRIMNATOR`. Technically, there can be multiple discriminator nodes (even though practically it's never the case), so if we encounter more than one, we just add `_2`, `_3`, etc as a suffix. Here, the value of this constant will simply be the `Uint8Array` represented by the node.
    - Finally, if it is a `SizeDiscriminatorNode`, we simply ignore it as it discriminate instruction via the size of their data and therefore is irrelevant here.
3. Use that new fragment in the instruction page.

That way, the type manifest doesn't need to be aware of these constant exports and we leave non-discriminator default values alone.

What do you think?

wjthieme:
Wondering what other type of `omitted` default values (apart from discriminators) there would be and if anyone would ever want access to these (if there are any). For doing GPAs arguably any part of the account data that is constant (a.k.a. and 'omitted' default value) would potentially be useful to have exposed. Potentially the same assumption holds for 'omitted' default values / constants in the instruction data.

> That way, the type manifest doesn't need to be aware of these constant exports and we leave non-discriminator default values alone.

Only thing I would potentially be concerned about is that `type manifest` would not know which default field values are discriminators and which are other types of default values. You would then potentially have the same constant show up as a magic value in the default values like so (but that is just a styling thing and wouldn't really matter since it is all generated code anyway):

```
export const ALLOCATE_INSTRUCTION_DATA_DISCRIMINATOR = 8;

...

export function getAllocateInstructionDataEncoder(): Encoder<AllocateInstructionDataArgs> {
  return transformEncoder(
    getStructEncoder([
      ['discriminator', getU32Encoder()],
      ['space', getU64Encoder()],
    ]),
    (value) => ({ ...value, discriminator: 8 })
  );
}

Ultimately I do not really care so much about the actual implementation, just about the outcome (= having the DISCRIMINATORS exposed since that is an immediate improvement/nice-to-have from our end).

lorisleiva:
My main issue with this PR is that it doesn't really fulfill the brief. The brief being: "I want discriminator constants to use in my GPA filters".

It doesn't fulfill that brief because:
- We are exporting more than discriminators using a non-predicable heuristic (I.e. omitted default values).
- We are not supporting the other valid type of discriminator nodes: `BytesDiscriminatorNode`s.
- We are exporting the non-encoded values which won't be very helpful when using them as GPA filters. E.g. the discriminator value is `42`, how do I know how that number is encoded (u8, u32, etc.)?


The solution I propose fulfils the brief because:
- We are only exporting discriminators as constants.
- We are supporting all types of applicable discriminators.
- We are exporting the whole byte array, not the decoded value.

Regarding making the default value use the discriminator constant, that's a code-gen optimisation but it's totally feasible. We just need to check if the default value we are rendering is a discriminator and if so, use the constant instead.

I'm happy to spin up a PR tomorrow to show you what I mean by all this.

wjthieme:
> It doesn't fulfill that brief because

Fair enough! Right now we would only be needing Anchor account discriminators and potentially instruction discriminators. These are byte arrays so we wouldn't really care about the encoding. Your point about encoding of things like `number` etc. makes sense!

> I'm happy to spin up a PR tomorrow to show you what I mean by all this.

Awesome! Looking forward to it (feel free to close this)

lorisleiva:
@wjthieme This is what I had in mind. https://github.com/kinobi-so/kinobi/pull/168

Let me know what you think. 🙏

PR #168, opened 2024-08-15
Export discriminator constants for instructions and accounts

lorisleiva:
This PR is an alternative solution to https://github.com/kinobi-so/kinobi/pull/163 that answers the brief: "We need discriminator constants exported for instructions and/or accounts in order to use them in memcmp filters".

It exports a `MY_INSTRUCTION_MY_DISCRIMINATOR` or `MY_ACCOUNT_MY_DISCRIMINATOR` constant containing the non-encoded value for that discriminator. It also exports a `getMyInstructionMyDiscriminatorBytes()` or `getMyAccountMyDiscriminatorBytes()` function that returns the encoded value for that discriminator (using the appropriate codecs).

Additionally, when rendering struct default values, it will identify if a default value is associated with a discriminator constant and use it when applicable instead of repeating the value.

### DIFF
--- a/.changeset/itchy-mirrors-enjoy.md
+++ b/.changeset/itchy-mirrors-enjoy.md
@@ -1,0 +1,5 @@
+---
+'@kinobi-so/renderers-js': patch
+---
+
+Export discriminator constants for instructions and accounts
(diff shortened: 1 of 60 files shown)


wjthieme:
Love it! This is by far a better solution than #163

PR #175, opened 2024-08-18
[0.22] Remove `importFrom` attribute from link nodes and resolvers

### LINKED ISSUE #92: Remove `hooked` and `importFrom` ability from link nodes
This is a proposal that needs further discussion before implementing.

---

Currently, the `importFrom` attribute of link nodes such as `DefinedTypeLinkNode` enable users to opt-out of defining the type using `TypeNodes` and instead grab them from a `hooked` folder or a named module (such as a package for JS or a crate for Rust using a dependency map).

Whilst this enables renderers to inject "custom types" into the generated code, it is a hole in the information tree and something that will affect explorer or anyone else requiring that information.

This issue propose that we no longer allow link nodes to "link outside of the tree" to avoid information loss.

As for enabling custom type injection for renderers, this could be achieved using special renderer options that lets the visitor know which defined type should be replaced by a custom one. This may even be a better outcome as it'll allow each renderer to decide if they want to inject custom code or not.

lorisleiva:
This PR removes the `importFrom` attribute from all nodes — i.e. all `LinkNodes` and the `ResolverValueNode` — which allowed nodes to be linked from custom directories/libraries without being present in the Kinobi IDL. This is because this feature is very renderer-centric and creates a gap in the information tree provided by Kinobi. See https://github.com/kinobi-so/kinobi/issues/92 for more information.

This PR keeps the ability for renderers to override certain links with custom directories/libraries but, instead, it does this by providing a new option to each renderer. This option allows us to create link override and tell the renderer to fetch that link from a different place.

This option has the following type:

```typescript
export type LinkOverrides = {
    accounts?: Record<string, string>;
    definedTypes?: Record<string, string>;
    pdas?: Record<string, string>;
    programs?: Record<string, string>;
    resolvers?: Record<string, string>;
};
```

For instance, say we have the following two defined types such that one links to the other:

```typescript
const node = programNode({
    definedTypes: [
        definedTypeNode({
            name: 'symbol',
            type: fixedSizeTypeNode(stringTypeNode('utf8'), 5),
        }),
        definedTypeNode({
            name: 'metadata',
            type: structTypeNode([
                structFieldTypeNode({ name: 'identifier', type: definedTypeLinkNode('symbol') }),
            ]),
        }),
    ],
    name: 'myProgram',
    publicKey: '1111',
});
```

When rendering a JavaScript client, it will default to importing the linked type from the generated `symbol.ts` file. However, you may provide the following option to your renderer to make it a `hooked` type and therefore customise its implementation for JavaScript only.

```typescript
kinobi.accept(
  renderJavaScriptVisitor("clients/js/src/generated", {
    linkOverrides: {
      definedTypes: { symbol: 'hooked' }
    },
  })
);
```

This `linkOverrides` option is available on all the renderers. You can either pass the same option to all of them, or customise which renderer should have link overrides.

Notice how this gives us the power to customise how to import links without impacting the information tree which may be used for other purposes — e.g. explorers trying to parse data.

Closes #92


buffalojoec (approved):
Looks great!!

We discussed a lot of this offline, but I had one more thought jump out while reviewing. Is there enough overlap between `dependencyMap` and `linkOverrides` where they could be merged into the same renderer configuration? Or are they too distinctly different?

Either way I'm good with the changes.

lorisleiva:
Thanks!

Yes it's not super clear (I'm hoping better documentation can help here in the near future) but `dependencyMap` is a more generic option that enables us to transform any "**abstract** import path" into a "**concrete** import path" such as a relative path, a JS library or a Rust crate (for the Rust renderer).

For instance, the JS renderer uses abstract import paths such as `solanaCodecsCore` to target the `@solana/codecs-core` library. But then it is up to the user if they want to map this to `@solana/codecs-core`, `@solana/web3.js` or any fork they might have.

Another example is the `hooked` abstract path which will, by default, resolve to `../../hooked` in JS but could be configured to point to a different directory.

On the other hand, the `linkOverrides` option simply tells the renderer not to import a certain link node from the generated path but from the provided **abstract** import path instead. That abstract import path will then be resolved using the `dependencyMap` like any other abstract path.

Is this clearer or did I make that more confusing? 😅

buffalojoec:
> Is this clearer or did I make that more confusing? 😅

Yep, that's a bit more clear, thanks! Ship it!

PR #204, opened 2024-09-07
[0.21] Fix: kinobi error does not use optional account instruction name but InstructionNode

aspnxdd:
I noticed that the error `KINOBI_ERROR__VISITORS__CANNOT_USE_OPTIONAL_ACCOUNT_AS_PDA_SEED_VALUE,` was printing `[object Object]` on the instruction name.

With this PR we log the instruction name instead of `[object Object]` since it was using the `InstructionNode` object.

Prev error: 
```
KinobiError: Cannot use optional account [adminWallet] as the [adminWallet] PDA seed for the [adminLpTokenAccount] account of the [[object Object]] instruction.
```

Current error:
```
KinobiError: Cannot use optional account [adminWallet] as the [adminWallet] PDA seed for the [adminLpTokenAccount] account of the [createBook] instruction.
```

### DIFF
--- a/.changeset/chilly-numbers-teach.md
+++ b/.changeset/chilly-numbers-teach.md
@@ -1,0 +1,5 @@
+---
+'@kinobi-so/errors': patch
+---
+
+Fix `KINOBI_ERROR__VISITORS__CANNOT_USE_OPTIONAL_ACCOUNT_AS_PDA_SEED_VALUE` instructionName string to actually print the instructionName.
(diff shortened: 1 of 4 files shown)


lorisleiva:
Hey, thanks for fixing the error message. However, could you please keep the `instruction` variable in the error context? This enables people to access the instruction node if they catch the error.

lorisleiva (approved):
Thanks!

PR #224, opened 2024-10-01
[0.21] Add an optional programAddress override to instruction builders of JS renderer

mcintyre94:
Currently when an instruction creator is called, it hardcodes the `programAddress` part of the program to the program that the client is being rendered for. This PR adds a new optional `config` parameter to instruction creators, which can be used to override the `programAddress`. For example:

```ts
import {
    getTransferCheckedInstruction,
} from '@solana-program/token';

const sendTokenInstruction = getTransferCheckedInstruction({
    source,
    destination,
    amount,
    authority,
   decimals,
   mint,
  }, { programAddress: TOKEN_22_PROGRAM_ADDRESS }
);
```

This is useful when there are multiple programs with compatible instructions, for example in the case of token and token22. In these cases, when dealing with these compatible instructions, you can now re-use the `token` client and just change the `programAddress` of the generated instructions.

### DIFF
--- a/packages/renderers-js/test/instructionsPage.test.ts
+++ b/packages/renderers-js/test/instructionsPage.test.ts
@@ -397,3 +397,27 @@ test('it can override the import of a resolver value node', async () => {
         someModule: ['myResolver'],
     });
 });
+
+test('it renders optional config that can override the program address', async () => {
+    // Given the following instruction
+    const node = programNode({
+        instructions: [
+            instructionNode({
+                accounts: [],
+                name: 'myInstruction',
+            }),
+        ],
+        name: 'myProgram',
+        publicKey: '1111',
+    });
+
+    // When we render it.
+    const renderMap = visit(node, getRenderMapVisitor());
+
+    // Then we expect an optional config parameter with an optional programAddress field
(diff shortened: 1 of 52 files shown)


packages/renderers-js/test/instructionsPage.test.ts
```diff
@@ -397,3 +397,27 @@ test('it can override the import of a resolver value node', async () => {
...
+    });
+
+    // When we render it.
+    const renderMap = visit(node, getRenderMapVisitor());
+
+    // Then we expect an optional config parameter with an optional programAddress field
+    // And we expect this to be used to override programAddress if it is set
+    await renderMapContains(renderMap, 'instructions/myInstruction.ts', [
+        'config?: { programAddress?: Address }',
+        'programAddress = config?.programAddress ?? MY_PROGRAM_PROGRAM_ADDRESS',
```
mcintyre94, line 421: I don't love this test, it feels a bit too literal. But I think it's probably the best we can do without actually executing the generated code.

lorisleiva (approved):
Perfect. Thank you! 🍺

PR #240, opened 2024-10-13
Add support for verbatimModuleSyntax with the umi renderer

4lve:
This adds a "fixer" for all imports so if they are PascalCase they get a type prefix which makes the renderer compatible with verbatimModuleSyntax

### DIFF
--- a/packages/renderers-js-umi/public/templates/sharedPage.njk
+++ b/packages/renderers-js-umi/public/templates/sharedPage.njk
@@ -2,7 +2,7 @@
 {% import "macros.njk" as macros %}
 
 {% block main %}
-import { AccountMeta, isSigner, Pda, publicKey, PublicKey, Signer, isPda } from '@metaplex-foundation/umi';
+import { type AccountMeta, isSigner, type Pda, publicKey, type PublicKey, type Signer, isPda } from '@metaplex-foundation/umi';
 
 /**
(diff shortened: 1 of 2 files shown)


lorisleiva (changes requested):
Hi there, thanks for tackling this in the Umi renderer but I think it'd be better if this implementation was consistent with the new web3.js renderer.

If you look at the `renderers-js` package, you'll see `type MyType` is supported and each type import must explicitly tell us it wants a type by important `"type MyType"`. Then, on the `ImportMap`, we ensure that if we receive both `"type MyType"` and `"MyType"`, then we only import the latter.

4lve:
Hmm okay, I tried this first, but some imports were generated directly from shared types, but I will look into it if I stumble onto any issues further down

4lve:
Btw, Is there any community discord or anything for this project/solana open source in general?

lorisleiva:
> Btw, Is there any community discord or anything for this project/solana open source in general?

We didn't have a dedicated place for these sort of conversations (more like dozens of small Telegram groups haha) but we do now! We added a channel directly under the "Dev Tooling" section of the "Solana Tech" server on Discord. Here's the link.

https://discord.com/channels/428295358100013066/1295768706806321264

4lve:
Cool!

PR #242, opened 2024-10-15
Add options to configure how traits are rendered in Rust

lorisleiva:
This PR adds the following `TraitOptions` configuration object to the Rust renderer allowing users to tweak how traits are rendered in their clients.

```ts
export type TraitOptions = {
    /** The default traits to implement for all types. */
    baseDefaults?: string[];
    /**
     * The default traits to implement for data enums only — on top of the base defaults.
     * Data enums are enums with at least one non-unit variant.
     */
    dataEnumDefaults?: string[];
    /**
     * The mapping of feature flags to traits.
     * For each entry, the traits will be rendered within a
     * `#[cfg(feature = "feature_name", derive(Traits))]` attribute.
     */
    featureFlags?: Record<string, string[]>;
    /** The complete trait overrides of specific types. */
    overrides?: Record<string, string[]>;
    /**
     * The default traits to implement for scalar enums only — on top of the base defaults.
     * Scalar enums are enums with no variants or only unit variants.
     */
    scalarEnumDefaults?: string[];
    /** The default traits to implement for structs only — on top of the base defaults. */
    structDefaults?: string[];
    /** Whether or not to use the fully qualified name for traits, instead of importing them. */
    useFullyQualifiedName?: boolean;
};
```

The default values for these options are:

```ts
export const DEFAULT_TRAIT_OPTIONS: Required<TraitOptions> = {
    baseDefaults: [
        'borsh::BorshSerialize',
        'borsh::BorshDeserialize',
        'serde::Serialize',
        'serde::Deserialize',
        'Clone',
        'Debug',
        'Eq',
        'PartialEq',
    ],
    dataEnumDefaults: [],
    featureFlags: { serde: ['serde::Serialize', 'serde::Deserialize'] },
    overrides: {},
    scalarEnumDefaults: ['Copy', 'PartialOrd', 'Hash', 'num_derive::FromPrimitive'],
    structDefaults: [],
    useFullyQualifiedName: false,
};
```

See readme changes for more details.

### DIFF
--- a/packages/renderers-rust/README.md
+++ b/packages/renderers-rust/README.md
@@ -65,10 +65,19 @@ Note that you must provide the fully qualified name of the traits you provide (e
 
 ```ts
 const traitOptions = {
-    baseDefaults: ['borsh::BorshSerialize', 'borsh::BorshDeserialize', 'Clone', 'Debug', 'Eq', 'PartialEq'],
+    baseDefaults: [
+        'borsh::BorshSerialize',
+        'borsh::BorshDeserialize',
+        'serde::Serialize',
+        'serde::Deserialize',
+        'Clone',
+        'Debug',
+        'Eq',
+        'PartialEq',
+    ],
     dataEnumDefaults: [],
     scalarEnumDefaults: ['Copy', 'PartialOrd', 'Hash', 'num_derive::FromPrimitive'],
-    structDefaults: ['serde::Serialize', 'serde::Deserialize'],
+    structDefaults: [],
 };
 ```
 
@@ -135,3 +144,5 @@ use serde::Deserialize;
 // With `useFullyQualifiedName` set to `true`.
 #[derive(serde::Serialize, serde::Deserialize)]
 ```
+
+Note that any trait rendered under a feature flag will always use the Fully Qualified Name in order to ensure we only reference the trait when the feature is enabled.
--- a/packages/renderers-rust/README.md
+++ b/packages/renderers-rust/README.md
@@ -86,3 +86,52 @@ const traitOptions = {
     },
 };
 ```
+
+### Feature Flags
+
+You may also configure which traits should be rendered under a feature flag by using the `featureFlags` attribute. This attribute is a map where the keys are feature flag names and the values are the traits that should be rendered under that feature flag. Here is an example:
+
+```ts
+const traitOptions = {
+    featureFlags: { fruits: ['fruits::Apple', 'fruits::Banana'] },
+};
+```
+
+Now, if at any point, we encounter a `fruits::Apple` or `fruits::Banana` trait to be rendered (either as default traits or as overridden traits), they will be rendered under the `fruits` feature flag. For instance:
+
+```rust
+#[cfg(feature = "fruits", derive(fruits::Apple, fruits::Banana))]
+```
+
+By default, the `featureFlags` option is set to the following:
+
+```ts
+const traitOptions = {
+    featureFlags: { serde: ['serde::Serialize', 'serde::Deserialize'] },
+};
+```
+
+### Using the Fully Qualified Name
+
+By default, all traits are imported using the provided Fully Qualified Name which means their short name will be used within the `derive` attributes.
+
+However, you may want to avoid importing these traits and use the Fully Qualified Name directly in the generated code. To do so, you may use the `useFullyQualifiedName` attribute of the `traitOptions` object by setting it to `true`:
+
+```ts
+const traitOptions = {
+    useFullyQualifiedName: true,
+};
+```
+
+Here is an example of rendered traits with this option set to `true` and `false` (which is the default):
+
+```rust
+// With `useFullyQualifiedName` set to `false` (default).
+use serde::Serialize;
+use serde::Deserialize;
+// ...
+#[derive(Serialize, Deserialize)]
+
+// With `useFullyQualifiedName` set to `true`.
+#[derive(serde::Serialize, serde::Deserialize)]
+```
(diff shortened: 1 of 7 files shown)


packages/renderers-rust/README.md
```diff
@@ -37,12 +37,112 @@ codama.accept(renderVisitor(pathToGeneratedFolder, options));
...
+```
+
+By default, the `featureFlags` option is set to the following:
+
+```ts
+const traitOptions = {
+    featureFlags: { serde: ['serde::Serialize', 'serde::Deserialize'] },
+};
+```
+
```
febo, line 122: I think we could also add a note to say that feature flags have to be present on the `Cargo.toml` as well.

febo (approved):
Looks great! Couple of micro nits on the comments/docs.

pinocchio-rs (approved):

febo (approved):

packages/renderers-rust/README.md
```diff
@@ -37,12 +37,114 @@ codama.accept(renderVisitor(pathToGeneratedFolder, options));
...
+        'serde::Deserialize',
+        'Clone',
+        'Debug',
+        'Eq',
+        'PartialEq',
+    ],
+    dataEnumDefaults: [],
+    scalarEnumDefaults: ['Copy', 'PartialOrd', 'Hash', 'num_derive::FromPrimitive'],
+    structDefaults: [],
+};
```
buffalojoec, line 81: I think there's a case to be made for the default being zero traits and zero scalar enum configurations, but I don't feel too strongly about it.

packages/renderers-rust/README.md
```diff
@@ -37,12 +37,114 @@ codama.accept(renderVisitor(pathToGeneratedFolder, options));
...
+| `traitOptions`                | [`TraitOptions`](#trait-options)                                                                                        | `DEFAULT_TRAIT_OPTIONS` | A set of options that can be used to configure how traits are rendered for every Rust types. See [documentation below](#trait-options) for more information.                                     |
+
+## Trait Options
+
+The Rust renderer provides sensible default traits when generating the various Rust types you client will use. However, you may wish to configure these traits to better suit your needs. The `traitOptions` attribute is here to help you with that. Let's see the various settings it provides.
+
+### Default traits
+
+Using the `traitOptions` attribute, you may configure the default traits that will be applied to every Rust type. These default traits can be configured using 4 different attributes:
```
buffalojoec, line 57: micro-micro-nit:

```suggestion
Using the `traitOptions` attribute, you may configure the default traits that will be applied to every generated Rust type. These default traits can be configured using 4 different attributes:
```

packages/renderers-rust/README.md
```diff
@@ -37,12 +37,114 @@ codama.accept(renderVisitor(pathToGeneratedFolder, options));
...
+    ],
+    dataEnumDefaults: [],
+    scalarEnumDefaults: ['Copy', 'PartialOrd', 'Hash', 'num_derive::FromPrimitive'],
+    structDefaults: [],
+};
+```
+
+### Overridden traits
+
+In addition to configure the default traits, you may also override the traits for specific types. This will completely replace the default traits for the given type. To do so, you may use the `overrides` attribute of the `traitOptions` object.
```
buffalojoec, line 86: Slick, I like it!

packages/renderers-rust/README.md
```diff
@@ -37,12 +37,114 @@ codama.accept(renderVisitor(pathToGeneratedFolder, options));
...
+    overrides: {
+        myCustomType: ['Clone', 'my::custom::Trait', 'my::custom::OtherTrait'],
+        myTypeWithNoTraits: [],
+    },
+};
+```
+
+### Feature Flags
+
+You may also configure which traits should be rendered under a feature flag by using the `featureFlags` attribute. This attribute is a map where the keys are feature flag names and the values are the traits that should be rendered under that feature flag. Here is an example:
```
buffalojoec, line 101: I like this even more!

buffalojoec (commented):
Post-humous review here!

I love the change. I think it's going to make working with the generated types a lot easier and less restrictive. A couple of thoughts from my side:

- _Technically_ speaking, what these configs are representing are traits that are _derived_ on the types. There's nothing stopping developers from implementing traits on generated types in a file outside of the `generated` dir. I guess this would just be a naming nit then.
- Some traits often rely on other traits being implemented on the type as well. for example, bytemuck's `Pod` requires `Copy`. I think it's okay to not really worry about this hiccup, since devs can just add the trait necessary and regenerate.
- Some traits like bytemuck's `Pod` also require `#[repr(C)]`. Even without bytemuck, some types may want to use this (like enums). Should we include a config for `repr`?

packages/renderers-rust/README.md
lorisleiva: We had to keep them like that in order to avoid introducing breaking changes. This produces code in the exact same way we were before. But I think these are sensible defaults for most clients, right?

lorisleiva:
> Some traits like bytemuck's Pod also require #[repr(C)]. Even without bytemuck, some types may want to use this (like enums). Should we include a config for repr?

@buffalojoec Would this be a blocker for you right now on some programs? If not, let's wait until more people ask for it. Otherwise, we can brainstorm a new `representation` option inside this `traitOptions` object asap.

packages/renderers-rust/README.md
buffalojoec: Sure, it doesn't really hurt to just let people change the base defaults anyway, and avoid the breaking change.

buffalojoec:
> > Some traits like bytemuck's Pod also require #[repr(C)]. Even without bytemuck, some types may want to use this (like enums). Should we include a config for repr?
> 
> @buffalojoec Would this be a blocker for you right now on some programs? If not, let's wait until more people ask for it. Otherwise, we can brainstorm a new `representation` option inside this `traitOptions` object asap.

Nope, not a blocker for me. I think if anyone tries to use the new trait configs to set up bytemuck traits then they'll hit the `repr` problem, but we can wait for that yep.

Dodecahedr0x:
Hello, I'm trying to use Codama with a pinocchio+bytemuck program and I'm hitting this issue of the generated code not having the `#[repr(C)]`

PR #246, opened 2024-10-17
Fix Rust rendering when account list is empty

febo:
### Problem

Currently, the Rust renderer will generate code with clippy warnings when the account list is empty since it will unnecessarily use the length of the accounts (`0` in this case) in calculations.

### Solution

Update the rendering logic to test if the account list length is greater than `0` before using the value.


lorisleiva (approved):
Perfect 👌

PR #250, opened 2024-10-21
Add `anchorTraits` option to Rust renderer

febo:
### Problem

The Rust renderer includes the implementation of Anchor traits by default under the `anchor` and `anchor-idl-build` features. This creates a problem when generating a client for a `solana-program` version not compatible with Anchor.

### Solution

Add an `anchorTraits` option to the Rust renderer to be able to toggle the implementation of Anchor traits. The default value is set to `true` to keep the current behaviour.


lorisleiva (approved):
Thank you!

PR #299, opened 2024-11-08
Add fetch function helpers to accounts page

wjthieme:
Fetch helper functions are currently included in the js client but not in the rust client. This PR adds similar fetch helper functions to the rust renderer.

I've added it behind a feature flag because it requires solana-client and solana-sdk crates which you don't want for CPIs.

Let me know what you think!


lorisleiva:
Oh interesting, thanks for that! @febo any thoughts about shipping this?

wjthieme:
@febo if you could look at this that would be great! One thing I am not 100% sure about is that with this change every generated client will have their own version of `DecodedAccount` and `MaybeAccount` which won't be interoperable.

febo:
> @febo if you could look at this that would be great! One thing I am not 100% sure about is that with this change every generated client will have their own version of `DecodedAccount` and `MaybeAccount` which won't be interoperable.

This is very nice! You are right that by having the `DecodedAccount` and `MaybeAccount` accounts on each generated client will make them "different" types. That will probably be annoying if you are using two different client together.

Perhaps what we could do is to define these types and maybe the logic for the `fetch_*` methods on a separate crate and use this crate on the generated clients. We could call the crate `codama-rpc` – @lorisleiva what do you think in having a repository for this?

wjthieme:
Also, I think it might be more useful to use nonblocking rpc client instead for the fetch functions (the caller can make an async function sync by blocking or using a runtime but not the other way around). If that sounds good I can make that change!

lorisleiva:
> Perhaps what we could do is to define these types and maybe the logic for the fetch_* methods on a separate crate and use this crate on the generated clients. We could call the crate codama-rpc – @lorisleiva what do you think in having a repository for this?

I 100% think we need to have an external crate that is being used by the generated code to avoid code repetition. But I don't think this is Codama. This the equivalent of the new Web3.js for the JavaScript renderer. We need something like "Web3.rs" for the Rust renderer. In other words, a Rust client framework that includes something akin to codecs and helpers functions like the ones introduced in this PR.

febo:
> > Perhaps what we could do is to define these types and maybe the logic for the fetch_* methods on a separate crate and use this crate on the generated clients. We could call the crate codama-rpc – @lorisleiva what do you think in having a repository for this?
> 
> I 100% think we need to have an external crate that is being used by the generated code to avoid code repetition. But I don't think this is Codama. This the equivalent of the new Web3.js for the JavaScript renderer. We need something like "Web3.rs" for the Rust renderer. In other words, a Rust client framework that includes something akin to codecs and helpers functions like the ones introduced in this PR.

Nice, this is a good idea. @wjthieme Would you like to create this repo with the helpers and we can then use it on the renderers? The helpers will probably be independent of the Codama client, like, you could use them standalone in any application.

Once this is available as a crate, we can update this PR to make use of that and avoid duplicating the types/logic.

wjthieme:
> The helpers will probably be independent of the Codama client

I'd be down to create this. Do you have a location in mind for the repo or doesn't really matter?

febo:
> > The helpers will probably be independent of the Codama client
> 
> I'd be down to create this. Do you have a location in mind for the repo or doesn't really matter?

I don't think it matters – happy to help with naming, etc.

lorisleiva:
FYI, @joncinque and I recently talked about moving away from `kaigan` crate in order to use a more encompassing crate as a "client framework" that can then be used by the generated code. It looks like there's some work overlap here so maybe we could set up a call to define/plan this?

wjthieme:
@lorisleiva moved the types to the `sharedPage`

wjthieme:
@lorisleiva added the changeset and found that the sharedPage was never rendered anywhere in the rust client. I proposed a fix but wasn't fully sure when it should be rendered or not. Let me know if my implementation is correct!

.changeset/orange-ravens-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+'@codama/renderers-rust': patch
+---
+
+Fix a bug where sharedPage was not rendered in rust client
```
lorisleiva, line 5: This changeset should be removed because it is a change that happened within the PR. Changesets are here to describe what the PR itself changes from the current default branch.

lorisleiva (approved):
Just waiting for CI to pass but looks good otherwise, thanks!

@febo are you happy with this?

.changeset/orange-ravens-sleep.md
wjthieme: Before this PR the sharedPage was never rendered so I figured I should add a changeset for it. See the change in packages/renderers-rust/src/getRenderMapVisitor.ts:273

febo: Yes, we are not using the types from the `sharedPage`. Most likely we stop rendering it but did not remove the actual page. You can see here that `RemainderVec` used is from the `kaigan` crate.

https://github.com/codama-idl/codama/blob/1ca91a141f3da09388801f3f516dae4c44336539/packages/renderers-rust/src/getTypeManifestVisitor.ts#L74-L78

wjthieme: Yep that is true that these types will not be compatible across clients. @lorisleiva and I figured that for now that would be fine since we don't have a place yet to put it in an external crate and moving it after the fact can be done non-breaking and is relatively low-lift. I believe there were open questions about replacing 'kaigan' with something else, etc.

febo: Cool, so if you guys are aware of the limitation, I am good with it. The only thing is to remove any other type that was already in that page. Those are not in use.

lorisleiva:
Thanks! Could you just re-run the tests locally and commit the changes so the e2e folders are updated as well?

febo:
The `sharedPage` looks good now. As Loris mentioned, we just need to update the examples.

packages/renderers-rust/e2e/dummy/src/generated/shared.rs
```diff
@@ -0,0 +1,10 @@
+//! This code was AUTOGENERATED using the codama library.
+//! Please DO NOT EDIT THIS FILE, instead use visitors
+//! to add features, then rerun codama to update it.
+//!
+//! <https://github.com/codama-idl/codama>
+//!
+
+
+
+
```
lorisleiva, line 10: Just checking, is Rust okay with importing empty modules?

wjthieme: I believe that should be fine but can change the condition to render the sharedPage if you think that is cleaner

febo: It should not generate an error, but it would be cleaner not to have an empty module.

wjthieme: Done!

lorisleiva (approved):
Thank you! 🍺

PR #314, opened 2024-11-22
Use program names when unwrapping link nodes

lorisleiva:
This PR fixes a few issues with the following three visitors.

- `unwrapInstructionArgsDefinedTypesVisitor` which uses:
- `unwrapDefinedTypesVisitor` which uses:
- `getDefinedTypeHistogramVisitor`

Namely, this PR makes it possible for identified `DefinedTypeLinkNodes` to be sandboxed by programs. That way, if two programs inside a `RootNode` provide a type node with the same name, unwrapping one will not affect the other.


febo (approved):
Nice! 👌

PR #336, opened 2024-11-28
Update usage of `IInstructionWithData`

lorisleiva:
This PR updates Codama and the code it generates for the new web3.js to be compatible with the latest `IInstructionWithData` change introduce in https://github.com/solana-labs/solana-web3.js/pull/3632.

Note that, we need to wait for `2.1.0` to be published before merging this PR since, otherwise, we will be using a canary version in our dependencies.

### DIFF
--- a/packages/renderers-js/public/templates/fragments/instructionParseFunction.njk
+++ b/packages/renderers-js/public/templates/fragments/instructionParseFunction.njk
@@ -33,7 +33,7 @@ export function {{ instructionParseFunction }}<
       & IInstructionWithAccounts<TAccountMetas>
     {% endif %}
     {% if hasData %}
-      & IInstructionWithData<Uint8Array>
+      & IInstructionWithData<ReadonlyUint8Array>
     {% endif %}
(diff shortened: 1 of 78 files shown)


mcintyre94:
Nice, thanks for this!

steveluscher:
Nice! I think this is the solution to the type error here, right? https://codesandbox.io/p/sandbox/dfsktf

packages/renderers-js/public/templates/fragments/instructionParseFunction.njk
```diff
@@ -33,7 +33,7 @@ export function {{ instructionParseFunction }}<
       & IInstructionWithAccounts<TAccountMetas>
     {% endif %}
     {% if hasData %}
-      & IInstructionWithData<Uint8Array>
+      & IInstructionWithData<ReadonlyUint8Array>
```
steveluscher, line 36: Nice! I think this is the solution to the type error here, right? https://codesandbox.io/p/sandbox/dfsktf

PR #348, opened 2024-12-05
Integrate `yellowstone-vixen-parser` with codama

Nagaprasadvr:
### Goal 
`yellostone-vixen-parser` crate helps with parsing accounts , instructions 
We can automate the process of generating these rust acc and ixs parsers using codama idl so anyone can add this as a plugin and parse their program  state and ixs effectively in rust

- [x] Add accounts_parser template
- [x] Add instructions_parser template
- [x] Add logic to parse codama idl and generate rust code for these parsers


packages/renderers-rust/public/templates/instructionsCpiPage.njk
```diff
@@ -150,9 +150,9 @@ impl<'a, 'b> {{ instruction.name | pascalCase }}Cpi<'a, 'b> {
           is_writable: remaining_account.2,
       })
     });
-    let {{ 'mut ' if hasArgs }}data = {{ instruction.name | pascalCase }}InstructionData::new().try_to_vec().unwrap();
+    let {{ 'mut ' if hasArgs }}data = borsh::to_vec(&{{ instruction.name | pascalCase }}InstructionData::new()).unwrap();
     {% if hasArgs %}
-      let mut args = self.__args.try_to_vec().unwrap();
+      let mut args = borsh::to_vec(&self.__args).unwrap();
```
lorisleiva, line 155: What's the rational for bumping the borsh crates here? IIRC, we went for the lowest possible version to avoid dependency hell as they are many libraries out there using different version of borsh. Cc @febo that might have a better memory than me here.

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
```diff
@@ -0,0 +1,213 @@
...
+        .replace(/\s+/g, '');
+};
+
+export const toSnakeCase = (str: string) => {
+    return str
+        .replace(/(?:^\w|[A-Z]|\b\w)/g, (word, index) => {
+            return index === 0 ? word.toLowerCase() : `_${word.toLowerCase()}`;
+        })
+        .replace(/\s+/g, '');
+};
```
lorisleiva, line 213: These helpers exist in `visitors-core`.

packages/renderers-rust/public/templates/instructionsCpiPage.njk
Nagaprasadvr: I was getting build errors about BorshDeserialize not being implemented

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
Nagaprasadvr: Will use them thanks

packages/renderers-rust/e2e/memo/Cargo.toml
```diff
@@ -11,7 +11,8 @@ test-sbf = []
 
 [dependencies]
 anchor-lang = { version = "0.30.0", optional = true }
-borsh = "^0.10"
+borsh = "^1.5.3"
```
febo, line 14: Same comment here about the `borsh` version.

packages/renderers-rust/public/templates/instructionsCpiPage.njk
Nagaprasadvr: switching to borsh 0.10

packages/renderers-rust/e2e/dummy/Cargo.toml
```diff
@@ -11,7 +11,8 @@ test-sbf = []
 
 [dependencies]
 anchor-lang = { version = "0.30.0", optional = true }
-borsh = "^0.10"
+borsh = "^1.5.3"
```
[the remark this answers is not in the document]
Nagaprasadvr: switching to borsh 0.10

packages/renderers-rust/e2e/memo/Cargo.toml
Nagaprasadvr: switching to borsh 0.10

kespinola (approved):

packages/renderers-rust/package.json
```diff
@@ -46,8 +46,10 @@
         "@codama/nodes": "workspace:*",
         "@codama/renderers-core": "workspace:*",
         "@codama/visitors-core": "workspace:*",
+        "@codama/nodes-from-anchor": "workspace:*",
         "@solana/codecs-strings": "rc",
-        "nunjucks": "^3.2.4"
+        "nunjucks": "^3.2.4",
+        "codama": "workspace:*"
```
lorisleiva, line 52: Hmm, any ideas why we need to top-level `codama` library here? This library should just be the final wrapper of all sub-libraries and therefore should never be needed inside the sub-libraries.

P.S.: if this is needed just to initialize some tests, we can use visitors directly on `RootNodes` without the need for the convenient `Codama` object.

packages/renderers-rust/package.json
```diff
@@ -46,8 +46,10 @@
         "@codama/nodes": "workspace:*",
         "@codama/renderers-core": "workspace:*",
         "@codama/visitors-core": "workspace:*",
+        "@codama/nodes-from-anchor": "workspace:*",
```
lorisleiva, line 49: I think this could/should be in `devDependencies`.

packages/renderers-vixen-parser/public/templates/programMod.njk
```diff
@@ -0,0 +1,8 @@
+{% extends "layout.njk" %}
+
+{% block main %}
+
+/// `{{ program.name | snakeCase }}` program ID.
+pub const {{ program.name | snakeCase | upper }}_ID: Pubkey = pubkey!("{{ program.publicKey }}");
+
+{% endblock %}
```
lorisleiva, line 8: Correct me if I'm wrong but I don't think this template is actually being used anywhere.

packages/renderers-vixen-parser/src/getTypeManifestVisitor.ts
lorisleiva, line 1: Just wanted to flag here that it could be a good idea in the future to have a `@codama/visitors-rust` helper package that provides Rust-specific helpers that can be reused in various Rust-based renderers so we don't have to duplicate the same visitors every time.

Nothing to change at this point but just something to keep in mind as we keep adding Rust renderers (e.g. the Pinocchio renderer in the works by @febo).

lorisleiva (changes requested):
I've made a few comments and we need to make sure CI passes before this can be merged but looks good overall! Also I think this needs a quick rebase from the main branch.

I don't understand the entire logic around the rendered parsers but one thing I noticed is it seems to discriminate all accounts by length instead of using the information provided by the `DiscriminatorNodes` in the standard. Not sure if that's by design at the moment but thought it needed to be flagged.

packages/renderers-rust/package.json
Nagaprasadvr: removed this

packages/renderers-vixen-parser/public/templates/programMod.njk
Nagaprasadvr: right , deleted

Nagaprasadvr:
there was an issue with rebase closing this and reopening a new pr

lorisleiva (changes requested):
CI not passing sorry.

lorisleiva (changes requested):
Just need to commit the generated changes in e2e tests now.

Nagaprasadvr:
> Just need to commit the generated changes in e2e tests now.

I don't see any local changes on my system

lorisleiva:
Strange, I also don't see any changes locally so it must be a CI thing. I'm trying to push commits to this branch to debug but it looks like I'm not allowed. Is there anything you can toggle to allow me to push to this PR?

lorisleiva:
We also need a changeset so we can trigger a publish of the new library on merge.

Nagaprasadvr:
> We also need a changeset so we can trigger a publish of the new library on merge.

Wil add

Nagaprasadvr:
where do we add the changeset?

lorisleiva:
I've just done it but, for next time, you just run `changeset` in the root directory locally and follow the prompts.

lorisleiva (approved):
Looks like we're good to merge! 👌

PR #380, opened 2025-01-09
renderer-rust: add decode-error trait to errors

buffalojoec:
Many Rust client applications may wish to define functions generic over `solana_program::decode_error::DecodeError`, which conveniently converts from a `ProgramError::Custom` to the respective enum value using `FromPrimitive`.

This change adds that trait implementation to the Rust renderer.


lorisleiva (approved):
Thank you!

PR #390, opened 2025-01-16
feat: support const program ids for pdas in anchor

lithdew:
Anchor allows for program ID's for PDA's to be specified using the 'seeds::program' parameter. Codama currently ignores these program ID's for PDA instruction accounts in Anchor when converting Anchor IDL nodes to Codama nodes. This commit implements support for constant program ID's for PDA instruction account nodes in Anchor.


.changeset/swift-tomatoes-care.md
```diff
@@ -0,0 +1,7 @@
+---
+'@codama/nodes-from-anchor': patch
+'@codama/renderers-js': patch
```
lorisleiva, line 3: Could you just remove this line from the changesets since this PR only includes test changes for this package.

lorisleiva (dismissed):
Awesome, Thanks for that! 🍺

.changeset/swift-tomatoes-care.md
lithdew: Ah yep - got that done.

lorisleiva (approved):
Thank you!

PR #401, opened 2025-01-23
bug: rust fetch return types not qualified properly

wjthieme:
introduced by #299. Tests did not catch this because the functions/types are behind a feature

* Fixed the bug with imports not being resolved
* Fixed an issue with errors from fetch functions not being converted 
* Added the fetch feature to the system example (with required optional dependencies)


febo (commented):
Looks good. A comment not directly related to this PR: would it make sense for the `fetch_all_*` variants to take a `Vec<&Pubkey>` since the `fetch_` ones take a `&Pubkey`?

wjthieme:
@febo I believe solana_client crate does `&Pubkey` for `get_account` and ` &[Pubkey]` for `get_multiple_accounts`. Would it make sense to match that?

(Changing would probably be fine given this has never worked in a live version even though it would strictly be a breaking change)

febo (approved):

febo:
> @febo I believe solana_client crate does `&Pubkey` for `get_account` and `&[Pubkey]` for `get_multiple_accounts`. Would it make sense to match that?
> 
> (Changing would probably be fine given this has never worked in a live version even though it would strictly be a breaking change)

Agreed, let's match the API of `solana_client`.

wjthieme:
> Agreed, let's match the API of solana_client.

Fixed!

lorisleiva:
Thank you! Just waiting on Febo's approval and LGTM.

Would you mind just rebasing the PR? Sorry I just merged another big PR and it created small conflicts.

lorisleiva (approved):

febo (approved):
Looks good, thanks!

PR #405, opened 2025-01-24
bug: serde serialize/deserialize big arrays

### LINKED ISSUE #184: [renderers-rust] Generated rust arrays longer than 32 do not serde `Serialize`
Arrays longer than 32 cannot be Serialized with serde directly (catch-32). I suspect that is why the `#[cfg_attr(feature = "serde", serde(with = "serde_with::As::<serde_with::Bytes>"))]` proc macro is added to array fields longer than 32.

This however still fails with the following error:
```
the trait bound `serde_with::Bytes: SerializeAs<[T; N]>` is not satisfied
```

Example: The TickArray account in https://github.com/orca-so/whirlpools/tree/main/rust-sdk/client when compiled with `serde` feature enabled. This account has a field `pub ticks: [Tick; 88]` which it fails on.

wjthieme:
resolves #184

`serde(with = "serde_with::As::<serde_with::Bytes>"))` does not work for arrays >32 items but is currently added. This causes a compiler error. Theoretically serde_with's docs state that `serde(with = "serde_with::As::<[_, X]>"))` should work but couldn't get it to compile. I introduced `serde-big-array` to resolve the issues.

Downside of this is that it adds an extra dependency (but only with `serde` feature enabled)

### DIFF
--- a/.changeset/tiny-bikes-hug.md
+++ b/.changeset/tiny-bikes-hug.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-rust': patch
+---
+
+bug: serde serialize/deserialize big arrays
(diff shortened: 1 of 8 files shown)


febo (approved):
Looks good, thanks!

PR #415, opened 2025-01-27
Update Lamports type from web3.js v2

lorisleiva:
Since we're now using the latest version of web3.js v2 (instead of `rc`), we need to adjust the `Lamports` type.


febo (approved):

lorisleiva:
### Merge activity

* **Jan 27, 11:47 AM EST**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/codama-idl/codama/415).

PR #438, opened 2025-02-06
Generate parsers for instructions with optional accounts

slimeonmyhead:
Not sure if I'm missing anything with this fix, but it works with the parsers that I tested it on.
The `check_min_accounts_req` is still in place at the top, so it will only parse if the account is defined.

My solana knowledge is lacking here: Is there a situation where an instruction has an optional account in the sdk and doesn't default to some value in the on-chain call?

In the cases that I checked (such as meteora swap and jupiter route) the optional accounts are defaulted to the program.


lorisleiva:
I am not really qualified to review the domain logic of this renderer so pinging @kespinola so he can take a look. 🙏

Nagaprasadvr (changes requested):
few changes to fix logic

Nagaprasadvr:
just checked rust sdk logic - which is defaulting to program id whn processing optional accounts , so we need to add checks for program id as well here

PR #440, opened 2025-02-08
renderers-rust: Use `borsh::to_vec`

### LINKED ISSUE #207: Rust renderer generates invalid code for newer borsh versions
The try_to_vec method was removed in [borsh v1.0](https://github.com/near/borsh-rs/blob/master/CHANGELOG.md#100-alpha6---2023-10-02)

Replace `x.try_to_vec().unwrap();` with `to_vec(&x).unwrap();`

febo:
### Problem

As described in #207 , `try_to_vec` was removed from `borsh` in versions `1.x`. This causes problems since the rust renderer generates code using it.

### Solution

Replace the use of `try_to_vec` by `borsh::to_vec`, which is available in versions of `borsh` < `1`.

Fixes #207

### DIFF
--- a/.changeset/nasty-dots-lie.md
+++ b/.changeset/nasty-dots-lie.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-rust': patch
+---
+
+Use borsh::to_vec in rust renderer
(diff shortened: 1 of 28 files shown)


lorisleiva (approved):
Very nice! 🍺

PR #452, opened 2025-02-11
fix: allow empty structs in v1 anchor to be parsed

stegaBOB:
Anchor will skip serializing the `fields` field in v1 structs if there are no fields. This change allows Codama to recognize those types as struct types. The logic to default the fields is already in place, so this just allows it to be used.

### DIFF
--- a/.changeset/fuzzy-nails-wait.md
+++ b/.changeset/fuzzy-nails-wait.md
@@ -1,0 +1,5 @@
+---
+'@codama/nodes-from-anchor': patch
+---
+
+recognize unit structs types from anchor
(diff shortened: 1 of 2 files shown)


lorisleiva:
Hey Sammy, thanks for this! I'm not sure I understand how this works though because this PR is making the if-statement condition stronger which means the use-case where `fields` doesn't exist is just gonna jump to the throw statement below, right?

P.S.: Could you add a `patch` changeset by running `changeset` locally? This will make sure a new version can be released from this PR.

stegaBOB:
Just added the changeset and got all the tests to pass! They were probably failing due to me being on a mac earlier. I think you may have misread the change. It's should be removing the `&& 'fields' in idlType` clause.

lorisleiva (approved):
You're totally right, thank you!

PR #455, opened 2025-02-12
Dynamically sized discriminator for anchor 0.31.0

tomjohn1028:
_This might be worth holding off on or putting into an anchor-0.31.x branch to avoid breaking changes. Would be nice to get a beta version with support released._

Relevant [Anchor change](https://github.com/coral-xyz/anchor/pull/3098)


lorisleiva:
Supporting multiple version of dependencies to generate the client is getting more and more hellish. I wonder if we need some logic that parses the appropriate `Cargo.toml` for the Rust client and figures out the version of each dependency to generate code that's compatible with these versions. Cc @febo.

PR #462, opened 2025-02-17
Add program link handling logic to linkOverrides

stegaBOB:
Introduce a helper function to handle program links in `definedTypeLinkNode`. This ensures program links are resolved correctly while preserving existing fallback behavior. TODO note added to evaluate and apply consistency across all similar node types.

### DIFF
--- a/packages/renderers-js/src/utils/linkOverrides.ts
+++ b/packages/renderers-js/src/utils/linkOverrides.ts
@@ -49,13 +49,20 @@ export function getImportFromFactory(
         resolvers: overrides.resolvers ?? {},
     };
 
+    // todo: figure out if this logic makes sense, and if so apply to all the other 'ProgramNode-able" nodes
+    const handleProgramLink = (node?: ProgramLinkNode): string | undefined => {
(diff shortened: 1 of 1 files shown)


packages/renderers-js/src/utils/linkOverrides.ts
```diff
@@ -49,13 +49,20 @@ export function getImportFromFactory(
         resolvers: overrides.resolvers ?? {},
     };
 
+    // todo: figure out if this logic makes sense, and if so apply to all the other 'ProgramNode-able" nodes
+    const handleProgramLink = (node?: ProgramLinkNode): string | undefined => {
+        if (!node) return;
+        const programName = node.name;
+        return linkOverrides.programs[programName];
```
lorisleiva, line 56: Hmm I'm not sure if this is what we want since we'll end up linking a "program" override on a "defined type" override.

stegaBOB: Ah yes I meant to do it in the reverse. The definedTypes override should be taking precedent here.

PR #485, opened 2025-02-27
Update Codama to use `@solana/kit@^2.1.0` instead of `@solana/web3.js`

steveluscher:


steveluscher:
@lorisleiva there's something messed up about the lint step. It's trying to reach into `packages/renderers-js/e2e/**/dist/` and I can't make it stop. There are new `@typescript-eslint/*` suppressions in the built code that are getting caught in the lint filters and failing the build.

lorisleiva (approved):
Thanks! 🚀

PR #510, opened 2025-03-17
Rust Renderer: `fetch` Fixes

buffalojoec:
Fixes a few compiler errors from the `fetch` API for account state, produced by the Rust Renderer.

* If `Pubkey` is not imported in the base module, it's not available to the `fetch` feature either. Like other account state methods, use the full path qualifier.
* Fix a few clippy nits about unnecessary borrows.

### DIFF
--- a/.changeset/short-showers-shout.md
+++ b/.changeset/short-showers-shout.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-rust': patch
+---
+
+Patch a few compile errors in the account `fetch` API.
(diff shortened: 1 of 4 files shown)


lorisleiva:
Good catch! The only slight annoyance with using the FQN here is that it won't go through the `ImportMap` which may affect which package you import `Pubkey` from.

But I don't think the `ImportMap` of the Rust renderer is clever enough right now to understand that certain imports may need to be feature flagged. So perhaps your workaround is the best solution for now.

Cc @febo lmk what you think.

buffalojoec:
> Good catch! The only slight annoyance with using the FQN here is that it won't go through the `ImportMap` which may affect which package you import `Pubkey` from.

Sure, but it looks like the other templates are littered with `solana_program::pubkey::Pubkey` as it is.

lorisleiva (approved):
Oh okay, if it is already an issue then we can address this in a future PR. Let's get this merged and published.

PR #514, opened 2025-03-18
Rust Renderer: add clippy ignore for arithmetic

buffalojoec:
#### Problem

If you use a Codama-generated client in a library, and you try to lint that library with a `+nightly` toolchain, it will throw an error. Clippy looks through the dependencies as well, and it will barf on these "arithmetic side effects" issues.

#### Summary of Changes

Simply add `#[allow(clippy::arithmetic_side_effects)]` to the list of preexisting clippy ignores on these methods.


agca52 (commented):
Thank you

PR #533, opened 2025-04-02
Vix 75 fix codama vixen account parser

Nagaprasadvr:
Problem 
- Currently vixen parser is using accounts len to match accounts and parse them but there could be issues with accounts having same length

Solution
- use account discriminators for anchor accounts and also check if acc disc exists for shank accounts and use them to match accounts and if we cannot use discs fallback to account length in the logic


packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
```diff
@@ -61,50 +113,55 @@ export function getRenderMapVisitor(options: GetRenderMapOptions = {}) {
...
-                            }
+                        // checks for discriminator
+                        let discriminator = getIxDiscriminatorFromInstructionNode(ix);
+
+                        if (discriminator && Array.isArray(JSON.parse(discriminator) as string[])) {
+                            ixDiscLen = Array.from(JSON.parse(discriminator) as string[]).length;
+                        }
+
+                        if (discriminator && ixDiscLen == 1) {
+                            discriminator = `[${discriminator}]`;
```
kespinola, line 160: We can't assume the discriminator that is 1 byte is shank. I think to start we can assume the discriminator is in the beginning of the byte list starting at index 0 but can be n length long.

I does look like you caccount for that in follow up checks and resets just setting default to 1. Please confirm.

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
```diff
@@ -61,50 +113,55 @@ export function getRenderMapVisitor(options: GetRenderMapOptions = {}) {
...
-                                    IX_DATA_OFFSET = Array.from(JSON.parse(value) as string[]).length;
-                                } else {
-                                    discriminator = `[${discriminator}]`;
-                                }
-                            }
+                        // checks for discriminator
+                        let discriminator = getIxDiscriminatorFromInstructionNode(ix);
+
+                        if (discriminator && Array.isArray(JSON.parse(discriminator) as string[])) {
+                            ixDiscLen = Array.from(JSON.parse(discriminator) as string[]).length;
```
kespinola, line 156: Assign to variable the Array.isArray(...)

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
Nagaprasadvr: yeah we just assumne the disc is 1 byte and later update that accordingly

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
Nagaprasadvr: done

kespinola (approved):

kespinola (approved):

lorisleiva (changes requested):
I'm a bit confused about the approach used here.

The Codama standard was specifically designed to be framework agnostic. It offers `DiscriminatorNodes` that describes how accounts and instructions are recognised regardless of how that program was created.

When a Codama IDL is given, having to go back to figuring out the program framework feels like a step back and it won't benefit from any further abstraction the standard provides in the future. Nor will it benefit from updates in adapters to and from other IDLs.

If you want to generate helpers that properly leverage the Codama IDL, you have to work with `DiscriminatorNodes` directly. If it is an Anchor program, the `DiscriminatorNodes` will already have been set by the `@codama/nodes-from-anchor` package. There is no need to compute the hashes again.

I strongly recommend looking into how the `@codama/renderers-js` package deals with `DiscriminatorNodes` to get an idea on how to implement this properly. For instance the following files could be a good starting point: [`discriminatorCondition.ts`](https://github.com/codama-idl/codama/blob/main/packages/renderers-js/src/fragments/discriminatorCondition.ts) and [`discriminatorConstants.ts`](https://github.com/codama-idl/codama/blob/main/packages/renderers-js/src/fragments/discriminatorConstants.ts).

fernandodeluret:
Hi @lorisleiva! Hope you are doing well. Thanks for your feedback! I've kept working over this in the context of adding support for generating the grpc server related part of the generated vixen parser. In general I have tried to follow your recomendations and now the discriminator processing is abstracted from if it was a shank or anchor idl (you can look at the basic support I've added here https://github.com/abklabs/codama/blob/vix-77-add-support-to-generate-impl-for-proto-structs-and-rust/packages/renderers-vixen-parser/src/getRenderMapVisitor.ts#L305-L366).

But the one thing I feel we are missing is a default discriminator generation for shank idls, (and please correct me if I'm missing something) because the `nodes-from-anchor` package is setting the discriminator nodes if it's an anchor idl with the default approach for anchor discriminators, but just leaving it empty for shank idls. 

So what I did so far is update the codama root object with the "common" approach for shank idls (1 byte at the start of the ix data for ixs and account length for accounts) in the generation script here https://github.com/abklabs/codama/blob/vix-77-add-support-to-generate-impl-for-proto-structs-and-rust/packages/renderers-vixen-parser/e2e/raydium-amm-v4-parser/codama.cjs#L157-L209. But that means that any user of the `renderers-vixen-parser` package and codama in general would have to manually set the discriminators in a similar fashion. So I would like to move that last small piece of code inside our packages for making it work out of the box for users if they don't want/need to customize anything (could be in `renderers-vixen-parser` because having that default it's very opinionated or also inside  `nodes-from-anchor`). Do you have any thoughts/recommendations regarding that?

Also as a side note, I'm finishing polishing a couple of edge cases in the proto generation and once that's done the plan is create a PR from that branch I shared to include all that logic (tests are already passing and the e2e tests have the format we have been using for the new set of parsers we have released in vixen these last days)

fernandodeluret:
Thanks for the response @lorisleiva! I've created both PRs:
- https://github.com/codama-idl/codama/pull/589 the big one with all the proto logic we have done so far
- https://github.com/codama-idl/codama/pull/590 the small one only adding default ix discriminators for shank idls

Also I pushed this branch using the changes from the discrminators PR for the shank example idl we had in our e2e tests and is working correctly: https://github.com/abklabs/codama/commit/1f4bac40a47bb3a8999298179bae74cba654ffeb

Nagaprasadvr:
closing this as we have a diff pr which covers this

lorisleiva:
Thanks guys, I'll take a look at the new PRs asap.

PR #589, opened 2025-05-08
Proto generation for Vixen Render

fernandodeluret:
## Overview
The render now generates code needed to support proto feature of Vixen. Simplification of end-user API.

## Changes
- Generate rust sdk within the vixen since always required and able to remove top level config. You can access the rust SDK from the Vixen crate.
- Support proto feature of Vixen. 
- Generate open telemetry traces for parsing
- Handle more variants of IDL (primarily shank flavor).
- Regenerate e2e parsers and re-test.
- Expand unit test coverage based on changes to the render.


packages/renderers-rust/e2e/dummy/idl.json
lorisleiva, line 1: Could you undo the indentation change of this file so it's easier for me to understand what changed?

Also, any particular reason you needed to change the `renderers-rust` e2e tests here? If this is needed, perhaps this could be in a separate PR as it's a bit out of scope here.

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
```diff
@@ -45,14 +70,320 @@ type ParserInstructionNode = {
...
+        case 'u8':
+        case 'u16':
+            return `.into()`;
+
+        default:
+            return ``;
+    }
+}
+
+// TODO!: Make this function use getTransform() for the inner items
```
lorisleiva, line 147: Just checking you didn't forget something here.

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
```diff
@@ -45,14 +70,320 @@ type ParserInstructionNode = {
...
+        throw new Error('Invalid discriminator length');
+    }
+
+    return discriminator;
+};
+
+/** Currenty only supports one discriminator per account, being in the discriminator field and at the start of the data.
+ * If not present or `sizeDiscriminator` is set, defaults to account LEN strategy.
+ */
+const getAccountSizeOrFieldDiscriminator = (node: AccountNode): string[] | null => {
```
lorisleiva, line 341: Any reason for supporting only field and size discriminators? Byte discriminators should be easy since it's just comparing a portion of bytes at the provided offset.

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
```diff
@@ -158,38 +524,246 @@ export function getRenderMapVisitor(options: GetRenderMapOptions = {}) {
...
+                            render('instructionsParserPage.njk', ixCtx),
+                        );
+                    }
+
+                    if (
+                        map.has(`src/${folderName}/accounts_parser.rs`) ||
+                        map.has(`src/${folderName}/instructions_parser.rs`)
+                    ) {
+                        // todo
                     }
```
lorisleiva, line 738: Did you forget something here?

lorisleiva (commented):
After a first pass on this PR, I can see this is a better approach than before. My main issue here is that this PR does way too much. It's tricky to review and make sure no bugs or misuse of the standard are being introduced.

Since this seems to add a new Codama to Protobuf generator, how would you feel about creating a `@codama/renderers-protobuf` package that only does this. Then you could use this package as a dependency for the vixen parser and only focus on the parts that are unique to vixen. Would that make sense or is the protobuf conversion too opinionated to be extracted as a generic tool?

packages/renderers-rust/e2e/dummy/idl.json
fernandodeluret: No no, you are right, we don't need to modify this here. I just rolled back all changes outside the renderers-vixen-parser package

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
fernandodeluret: Thanks for noticing, no I didn't miss it, it's a todo I wanted to leave there because it feels like the correct way to implement it. Actually if you see the other kind of "recursive" cases we implemented them use that getTransform function. But in this case it was a bit more trickier, we actually don't support more than 1 depth of nested arrays for the proto generation yet for example.

packages/renderers-vixen-parser/src/getRenderMapVisitor.ts
fernandodeluret: Hi @lorisleiva! I can answer here to this and also to your general concern about the PR because it's pretty related. I agree that this PR is quite big, that's not ideal I know. But we wanted to contribute the changes so far before it becomes even bigger and more complex. Because the other half of that sentence is that this is a first working iteration on the feature, but this is by no means feature complete.

The general approach we took is just be adding/improving implementations when we find an idl that presents a use case that we can't handle, and try to be as taxative as possible with the cases we support, and just rise an error if we don't yet (I think there is only one place we are still generating code out of a default switch branch instead of erroring).

This last part is also valid for the proto file generation. We always planned it only for use in the vixen parser no as a generic proto generation. It might turn out that is not that bad at fitting that role, but I'm not sure how much limited it can be to be honest. (If you ask me, personally I would prefer to leave as it is inside the package at least until we feel more confident about it, when this whole scenario is more polished and closer to be feature complete)

With that context, this is just one more of the many things not implemented yet. We only added support for field discriminators because are the more common ones generated by nodes-from-anchor default case and only at the start of the instruction data or account data, but for sure at some point we are going to need to add support for other types of discriminators and mainly for offsets

fernandodeluret: That said, I'll try to add today some polish to the package API, mainly the goal is to simplify it, so that the codama.cjs script that we have in each e2e example becomes simpler and with simpler configurations params. That might help reduce a bit some cyclomatic complexity from the package (it's not going to make a huge difference I know, but at least it will leave us with a more stable API for the users that we don't need to break inmediatly after)

fernandodeluret: I just pushed the clean up of the API I mentioned above

packages/renderers-vixen-parser/e2e/raydium-amm-v4-parser/codama.cjs
```diff
@@ -0,0 +1,175 @@
...
+    const updatedNode = { ...node, program: { ...node.program, definedTypes, accounts } };
+
+    visit(
+        updatedNode,
+        renderVisitor({
+            projectFolder: __dirname,
+            projectName: project,
+        }),
+    );
+
```
kespinola, line 103: @fernandodeluret that is a nice looking top level API. Thank you for making these changes.

kespinola (approved):
> Since this seems to add a new Codama to Protobuf generator, how would you feel about creating a @codama/renderers-protobuf package that only does this. Then you could use this package as a dependency for the vixen parser and only focus on the parts that are unique to vixen. Would that make sense or is the protobuf conversion too opinionated to be extracted as a generic tool?

@lorisleiva for now the proto is an extension of tailor for Vixen stream feature. There are likely pieces that could be isolated but don't want to split out too early. It is cool that this generates a proto def for the IDL that could be useful in other situations maybe on a grpc based RPC stack that returns parsed results.


Overall, @lorisleiva all these changes are isolated to the renderers-vixen-parser and we have 10+ new parsers based on these updates so we are happy with its capabilities right now and would like to get a release out so we can update community on updated generation.

lorisleiva:
Thanks for the update Kyle, I'm happy to get that merged if you guys are happy with the PR. Just need to get CI to green first and I'll approve. 🙏

fernandodeluret:
> Thanks for the update Kyle, I'm happy to get that merged if you guys are happy with the PR. Just need to get CI to green first and I'll approve. 🙏

Thanks @lorisleiva! :pray:  This is done, just updated the branch and the tests. Also now that we have the shank discriminators merged in, I updated the e2e script here to showcase that here https://github.com/codama-idl/codama/pull/589/commits/eeb01989a2c5e92bbdf190180c5ca854af1e4aab#diff-b9cdaef35149142caeaf31a6606a6891e1c88d37cb45e2c240549810487c0fa9

packages/renderers-vixen-parser/public/templates/libPage.njk
```diff
@@ -0,0 +1,20 @@
+{% block main %}
+mod generated_parser;
+mod generated_sdk;
+pub use generated::*;
+pub use generated_parser::*;
+use generated_sdk as generated;
+use solana_program::pubkey::Pubkey;
```
sonicfromnewyoke, line 7: i'd recommend getting rid of `solana_program` in favour of using direct crates. It should drastically improve compilation speed
```suggestion
use solana_pubkey::Pubkey;
```

packages/renderers-vixen-parser/public/templates/CargoPage.njk
```diff
@@ -0,0 +1,34 @@
...
+[package]
+name = "{{ projectName }}"
+version = "0.1.0"
+edition = "2021"
+
+
+[dependencies]
+prost = "0.13.1"
+tonic = { version = "0.12.1", features = ["gzip", "zstd"] }
+solana-program = "^2.1.6"
```
sonicfromnewyoke, line 13: i believe we are  already ready to use `~2.2` version of solana deps, as this version is recommended right now
```suggestion
solana-pubkey = "~2.2"
```

also, maybe it makes sense to use 
```toml
solana-pubkey = { workspace = true }
```
to benefit from rust workspaces

packages/renderers-vixen-parser/public/templates/CargoPage.njk
```diff
@@ -0,0 +1,34 @@
...
+borsh = "^0.10"
+yellowstone-vixen-core = { git = "https://github.com/rpcpool/yellowstone-vixen", branch = "main",features = ["proto"] }
```
sonicfromnewyoke, line 15: wdyt about using workspace here?
```suggestion
yellowstone-vixen-core = { workspace = true, features = ["proto"] }
```

packages/renderers-vixen-parser/e2e/orca-whirlpool-parser/src/generated_sdk/accounts/fee_tier.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub const LEN: usize = 44;
+
+    #[inline(always)]
+    pub fn from_bytes(data: &[u8]) -> Result<Self, std::io::Error> {
+        let mut data = data;
+        Self::deserialize(&mut data)
+    }
+}
+
+impl<'a> TryFrom<&solana_program::account_info::AccountInfo<'a>> for FeeTier {
```
sonicfromnewyoke, line 35: i'd recommend using `solana-account-info` instead

packages/renderers-vixen-parser/e2e/orca-whirlpool-parser/src/generated_sdk/instructions/set_collect_protocol_fees_authority.rs
```diff
@@ -0,0 +1,391 @@
...
+pub struct SetCollectProtocolFeesAuthority {
+    pub whirlpools_config: solana_program::pubkey::Pubkey,
+
+    pub collect_protocol_fees_authority: solana_program::pubkey::Pubkey,
+
+    pub new_collect_protocol_fees_authority: solana_program::pubkey::Pubkey,
+}
+
+impl SetCollectProtocolFeesAuthority {
+    pub fn instruction(&self) -> solana_program::instruction::Instruction {
```
sonicfromnewyoke, line 22: i'd recommend using `solana-instruction` here

packages/renderers-vixen-parser/e2e/orca-whirlpool-parser/src/generated_sdk/instructions/set_collect_protocol_fees_authority.rs
```diff
@@ -0,0 +1,391 @@
...
+    ) -> Self {
+        Self {
+            __program: program,
+            whirlpools_config: accounts.whirlpools_config,
+            collect_protocol_fees_authority: accounts.collect_protocol_fees_authority,
+            new_collect_protocol_fees_authority: accounts.new_collect_protocol_fees_authority,
+        }
+    }
+    #[inline(always)]
+    pub fn invoke(&self) -> solana_program::entrypoint::ProgramResult {
```
sonicfromnewyoke, line 188: i'd recommend using `solana-program-entrypoint` here

sonicfromnewyoke (changes requested):
overall looks great to me.

if you don't mind, implementing the next changes, which are described in this [solana-sdk-crate-migration-guide](https://github.com/kevinheavey/solana-sdk-crate-migration-guide/blob/main/README.md)

this migration should drastically reduce compilation time and resulting binary size

packages/renderers-vixen-parser/public/templates/CargoPage.njk
fernandodeluret: I'll answer here because this is the most anoying one, but this actually apply for all workspaces related comments. In general I think we can't assume that the crate is going to be used inside a workspace, i.e. if we add this change here we break all the e2e tests because there is no workspace there.

I'll start taking a better look at the rest of your comments in a moment.

packages/renderers-vixen-parser/package.json
```diff
@@ -43,12 +43,14 @@
...
         "@codama/visitors-core": "workspace:*",
+        "@codama/renderers-rust": "workspace:*",
         "@solana/codecs-strings": "rc",
         "nunjucks": "^3.2.4"
     },
     "devDependencies": {
         "@types/nunjucks": "^3.2.6",
-        "@codama/nodes-from-anchor": "workspace:*"
+        "@codama/nodes-from-anchor": "workspace:*",
+        "@codama/visitors": "workspace:*"
```
lorisleiva, line 53: nit: I could be wrong but I couldn't see any new imports that require this new dev dependency.

lorisleiva (approved):
Thanks! I can see there are some ongoing comments but feel free to merge whenever you're happy with the state of the PR.

packages/renderers-vixen-parser/package.json
fernandodeluret: No no, you are right! (I think we might have been using it at some point in the e2e tests and then forgot to remove the devDep). Thanks for pointing out. I'll update this and fix the lock file conflicts :handshake:

PR #590, opened 2025-05-09
Add support for default shank ix discriminators in nodes-from-anchor

fernandodeluret:
This PR adds a default instruction discriminator for Shank idls, the generated value is a 1 byte discriminator at the start of the ix data based on the ix index value on the idl

### DIFF
--- a/packages/nodes-from-anchor/README.md
+++ b/packages/nodes-from-anchor/README.md
@@ -15,8 +15,10 @@ This package converts Anchor IDLs from various versions into Codama IDLs.
 pnpm install @codama/nodes-from-anchor
 ```
 
-> [!NOTE]
-> This package is **not** included in the main [`codama`](../library) package.
+> ⚠️ NOTES ⚠️
+>
+> - This package is **not** included in the main [`codama`](../library) package.
+> - A fair word of advice is that in case the `metadata.origin` key of the idl is not set, it is assumed to be `"anchor"`. So if you are trying to parse a Shank IDL, be sure that origin is set to `"shank"`.
 
 ## Functions
 
--- a/packages/nodes-from-anchor/README.md
+++ b/packages/nodes-from-anchor/README.md
@@ -15,10 +15,10 @@ This package converts Anchor IDLs from various versions into Codama IDLs.
 pnpm install @codama/nodes-from-anchor
 ```
 
-> ⚠️ NOTES ⚠️
+> [!NOTE]
 >
 > - This package is **not** included in the main [`codama`](../library) package.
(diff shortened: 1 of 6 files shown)


packages/nodes-from-anchor/README.md
```diff
@@ -15,8 +15,10 @@ This package converts Anchor IDLs from various versions into Codama IDLs.
 pnpm install @codama/nodes-from-anchor
 ```
 
-> [!NOTE]
-> This package is **not** included in the main [`codama`](../library) package.
+> ⚠️ NOTES ⚠️
+>
+> - This package is **not** included in the main [`codama`](../library) package.
+> - A fair word of advice is that in case the `metadata.origin` key of the idl is not set, it is assumed to be `"anchor"`. So if you are trying to parse a Shank IDL, be sure that origin is set to `"shank"`.
```
lorisleiva, line 21: nit: `[!NOTE]` is a special MD notation that displays the message is a special colour on GitHub so best to keep it that way.

```suggestion
> [!NOTE]
>
> - This package is **not** included in the main [`codama`](../library) package.
> - If `metadata.origin` is not set on the IDL, it is assumed to be `"anchor"`. If you are trying to parse a Shank IDL, be sure that origin is set to `"shank"` so discriminators can be set correctly.
```

packages/nodes-from-anchor/src/v00/InstructionNode.ts
```diff
@@ -41,6 +46,15 @@ export function instructionNodeFromAnchorV00(idl: IdlV00Instruction, origin?: 'a
...
+    } else if (origin === 'shank') {
+        const discriminatorField = instructionArgumentNode({
+            defaultValue: bytesValueNode('base16', ixIndex.toString(16)),
+            defaultValueStrategy: 'omitted',
+            name: 'discriminator',
+            type: fixedSizeTypeNode(bytesTypeNode(), 1),
+        });
+        dataArguments = [discriminatorField, ...dataArguments];
+        discriminators = [fieldDiscriminatorNode('discriminator')];
     }
```
lorisleiva, line 58: I was going to comment saying that Shank actually [supports exporting custom discriminants for instructions](https://github.com/metaplex-foundation/shank/blob/46cbde90d46296c5c31660a8f6220817444e3921/shank-idl/src/idl_instruction.rs#L131-L144) and that we should support that first and use your solution as a fallback but scrolling up a bit, I can see we're already doing this in the first if statement of this function.

lorisleiva (approved):
Looks good! Just a little nit comment on the README but good to merge otherwise.

packages/nodes-from-anchor/src/v00/InstructionNode.ts
fernandodeluret: Yep, you are correct about it! :handshake:

kespinola (approved):

Issue #591, opened 2025-05-10
error[E0599]: no method named `as_bytes` found for array `[{integer}; 6]` in the current scope


ddldone:
```
pub const PREFIX: &'static [u8] = [99, 111, 110, 102, 105, 103].as_bytes();

pub fn create_pda(
    bump: u8,
) -> Result<solana_program::pubkey::Pubkey, solana_program::pubkey::PubkeyError> {
    solana_program::pubkey::Pubkey::create_program_address(
        &[[99, 111, 110, 102, 105, 103].as_bytes(), &[bump]],
        &crate::PROGRAM_CHAIN_ID,
    )
}

pub fn find_pda() -> (solana_program::pubkey::Pubkey, u8) {
    solana_program::pubkey::Pubkey::find_program_address(
        &[[99, 111, 110, 102, 105, 103].as_bytes()],
        &crate::PROGRAM_CHAIN_ID,
    )
}
```

lorisleiva:
Hi there! I'm afraid I'm not sure what this change is fixing and CI fails in the provided PR. Please re-submit with detailed information on the issue if you think there is still something to fix here.

PR #592, opened 2025-05-10
fix: method error of seeds

ddldone:


lorisleiva:
See https://github.com/codama-idl/codama/issues/591#issuecomment-2970887726

PR #596, opened 2025-05-12
Decouple the CLI from the main library

lorisleiva:
This is causing renderer updates to unnecessarily bump core Codama libraries.


febo (approved):
👌

PR #616, opened 2025-05-22
feat: replace solana-program with pubkey/account-info/instruction crates

sonicfromnewyoke:
## Problem
`renderers-rust` and `renderers-vixen-parser` use bloated `solana-sdk` and `solana-program` crates

## Solution
replace `solana-sdk` and `solana-program` by smaller creates which leads to faster compilation time and smaller bin sizes according to [solana-sdk-crate-migration-guide](https://github.com/kevinheavey/solana-sdk-crate-migration-guide/tree/main)

also `solana-*` crates were bumped to the `~v2.2`


fernandodeluret (approved):
From the perspective of `renderers-vixen-parser` (and the use we do there of `renderers-rust`) this is LGTM!  :handshake: :heart:

kespinola (approved):

lorisleiva (changes requested):
Thanks, we’ve actually already got this done for `solana-pubkey` (https://github.com/codama-idl/codama/pull/535) in a dedicated branch (`renderers-rust/use-granular-crates`) in the hope to tackle more granular crates as they get released. This is mainly so we can do a single breaking change for the Rust renderer for all these changes. Would you be able to submit multiple PRs for the other crates into this branch? Then we can publish the whole branch when it’s ready.

PR #644, opened 2025-06-12
[cli] correct readme commands

nickfrosty:
Since the codama cli is installed as a package into the user's repo (and not globally), they will need to use their package manager to run the codama cli

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -15,7 +15,7 @@ To get started with Codama, simply install `@codama/cli` which provides the `cod
 
 ```sh
 pnpm install @codama/cli
-codama init
+pnpm codama init
 ```
 
 You will be prompted for the path of your IDL and asked to select any script presets you would like to use.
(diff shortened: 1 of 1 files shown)


lorisleiva (commented):
Hmm I don't think the `pnpm` prefix is necessary — even when installing a package locally. The only advantage of using a `pnpm` prefix is that it will find the closest `package.json` wherever your are in the folder structure whereas using `codama` directly requires you run this at the root of the repository.

nickfrosty:
in my testing, if you dont add the prefix the `codama` command is not available since i do not have it installed globally or linked locally for maintainer testing.

do you not experience the same?

lorisleiva:
Oh I think it's probably because you don't have `./node_modules/.bin` in your local `PATH` variable. Maybe it's safer to document it as `pnpm codama` but it's a bit uglier haha.

lorisleiva:
Alright let's add the prefix then. 🫡

Could you please just remove the changeset since this PR does not warrant a new version to be published?

nickfrosty:
changeset removed :)

lorisleiva (approved):
Thank you!

PR #645, opened 2025-06-12
refactor(renderers-rust): extract `solana-account-info` from `solana-program` & `solana-account` from `solana-sdk`

sonicfromnewyoke:
### Problem

renderers-rust use bloated `solana-sdk` and `solana-program` crates

### Solution

extract `solana-account-info` from `solana-program`  and `solana-account` from `solana-sdk` as a part of more granular migration, agreed in #616


lorisleiva (approved):
Tysm! 🍻

PR #651, opened 2025-06-17
Renderers py

daog1:
This package provides Python client renderers for Codama IDLs.
It is fully operational and compatible with the majority of Codama node types. Test cases align with those in renderers-js.
Sample demonstrations of generated effects are available at https://github.com/daog1/PyGenIDL001.


lorisleiva (changes requested):
Sorry for the late review. This is looking promising. I've made a few high-level comments. I've not done any python for a while and never used the python SDKs for Solana, so I'd love for someone experiences with that world to give a deep review (@sonicfromnewyoke perhaps that's you?).

My overall feeling right now is that this is probably good enough but does not support the full type system that Codama describe. I can see this with the `extension` file that tries to patch specific use-cases. If that's the case, it's probably still okay to ship. The Rust renderer itself doesn't fully support the Codama type system because of the lack of a codec-like library (as Kit offers).

My most important question for whoever ends up shipping this Python render is: are you able to stick around once it's ship to continue maintaining it?

daog1:
> Sorry for the late review. This is looking promising. I've made a few high-level comments. I've not done any python for a while and never used the python SDKs for Solana, so I'd love for someone experiences with that world to give a deep review (@sonicfromnewyoke perhaps that's you?).很抱歉评论晚了。这看起来很有希望。我已经发表了一些高层次的评论。我已经有一段时间没有做过任何 python，也从未使用 Solana 的 python SDK，所以我希望有人对那个世界有经验的人进行深入的评论（也许就是你？
> 
> My overall feeling right now is that this is probably good enough but does not support the full type system that Codama describe. I can see this with the `extension` file that tries to patch specific use-cases. If that's the case, it's probably still okay to ship. The Rust renderer itself doesn't fully support the Codama type system because of the lack of a codec-like library (as Kit offers).我现在的总体感觉是，这可能已经足够好了，但不支持 Codama 描述的完整类型系统。我可以通过尝试修补特定用例的`扩展`文件看到这一点。如果是这样的话，可能还是可以发货的。Rust 渲染器本身并不完全支持 Codama 类型系统，因为缺少类似编解码器的库（如 Kit 提供的那样）。
> 
> My most important question for whoever ends up shipping this Python render is: are you able to stick around once it's shipped to continue maintaining it?对于最终发布此 Python 渲染的人，我最重要的问题是：一旦发布，您能否留下来继续维护它？

No problem, it will be maintained continuously.

lorisleiva:
Just to keep a record on GitHub about what has been discussed, the plan is now to have renderers has part of their own repositories, ideally owned by the team that maintains them. I have written a [more detailed explanation here](https://github.com/codama-idl/codama/issues/830#issuecomment-3271328972).

Regarding the python renderer, I can see it has already been moved to [its own repo available here](https://github.com/Solana-ZH/codama-py) meaning this PR can now be closed. Thank you so much for your work and I'll be sure to link this repository in the main README of the Codama repo.

PR #668, opened 2025-06-26
add gill support to the cli `init` command

nickfrosty:
added the ability for the codama cli to initialize a gill based codama config file using the `createCodamaConfig` function in gill. see gill docs here: https://gill.site/docs/guides/codama#create-a-codama-config

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -10,33 +10,40 @@ export function setInitCommand(program: Command): void {
         .argument('[output]', 'Optional path used to output the configuration file')
         .option('-d, --default', 'Bypass prompts and select all defaults options')
         .option('--js', 'Forces the output to be a JavaScript file')
+        .option('--gill', 'Forces the output to be a gill based JavaScript file')
         .action(doInit);
 }
 
 type InitOptions = {
     default?: boolean;
     js?: boolean;
+    gill?: boolean;
 };
+
+type ConfigFileType = "json" | "js" | "gill";
 
 async function doInit(explicitOutput: string | undefined, options: InitOptions) {
     const output = getOutputPath(explicitOutput, options);
-    const useJsFile = options.js || output.endsWith('.js');
+    let configFileType : ConfigFileType = output.endsWith(".json") ? "json" : "js";
+    if (options.gill) configFileType = "gill"
+    else if (options.js) configFileType = "js"
+
     if (await canRead(output)) {
         throw new Error(`Configuration file already exists at "${output}".`);
     }
 
     logBanner();
     const result = await getPromptResult(options);
-    const content = getContentFromPromptResult(result, useJsFile);
+    const content = getContentFromPromptResult(result, configFileType);
     await writeFile(output, content);
     logSuccess(`Configuration file created at "${output}".`);
 }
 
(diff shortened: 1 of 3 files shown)


lorisleiva (changes requested):
Thanks! Just a couple of points. 🙏

packages/cli/src/commands/init.ts
```diff
@@ -130,8 +137,16 @@ function getContentFromPromptResult(result: PromptResult, useJsFile: boolean): s
...
+    if (configFileType == 'json') {
         return JSON.stringify(content, null, 4);
+    } else if (configFileType == 'gill') {
+        return (
+            `import { createCodamaConfig } from "gill";\n\n` +
+            `export default createCodamaConfig({ \n\t` +
+            `idl: "${result.idlPath}", \n\t` +
+            `clientJs: "${result.jsPath}", \n` +
+            `});`
+        );
```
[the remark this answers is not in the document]
nickfrosty: I just updated to include the rust client path. The gill function has been updated and will go live in the next release: https://github.com/solana-foundation/gill/pull/165

packages/cli/src/commands/init.ts
```diff
@@ -130,8 +137,16 @@ function getContentFromPromptResult(result: PromptResult, useJsFile: boolean): s
...
+    if (configFileType == 'json') {
         return JSON.stringify(content, null, 4);
+    } else if (configFileType == 'gill') {
+        return `import { createCodamaConfig } from "gill";\n\n` +
+            `export default createCodamaConfig({ \n\t` +
+            `idl: "${result.idlPath}", \n\t` +
+            `clientJs: "${result.jsPath}", \n` +
+            result.scripts.includes('rust')
+            ? `clientRust: "${result.rustPath}", \n`
+            : `` + `});`;
```
lorisleiva, line 149: Thank you for including the Rust client in there!

There's one more thing to consider. The JS client is also optional. So if the user only wants a Rust renderer, we shouldn't display the `clientJs` option on there. Is this currently supported by Gill?

lorisleiva (changes requested):

packages/cli/src/commands/init.ts
nickfrosty: gill's assumption is that if a user is using the `createCodamaClient` function they want a js client that works with gill (since they have to have it installed). So the `clientJs` argument is required.

lorisleiva: Gotcha, then perhaps we shouldn't allow the user not to select the JS client when using the `--gill` option?

nickfrosty: yeah, I think that's a better approach. They will still need to input the js client data. I think it would be the best experience if users that passed the `--gill` flow see the same "select which client" buttons, but with the js one locked as selected/on with a note saying something to the effect of "you are using gill, so you get a js client"

lorisleiva: Hey @nickfrosty, is this something you'd want to tackle in this PR? Or should we just merge this as-is for now and you can add that safety guard later on for Gill?

lorisleiva (approved):
Thanks!

packages/cli/src/commands/init.ts
lorisleiva: Sounds good! Issue created here: https://github.com/codama-idl/codama/issues/726.

PR #676, opened 2025-07-01
Update renderers-js to be compatible with ESM 

macalinao:
The existing version of renderers-js does not export code compatible with NodeNext module resolution, preventing use of `"type": "module"` with the latest TypeScript targets.

This patch adds this compatibility in addition to fixing several common linter and compile errors.


lorisleiva (changes requested):
Hi there, thanks for this but at the very least this should be an opt-in option on the renderer and not the default behaviour.

When exporting a library for both ESM and CJS (like we do in Solana Kit for instance) it is recommended not to use `"type": "module"` attribute and use CJS by default so type definitions are more portable across both ESM and CJS consumers.

This is why I believe the default behaviour should stay the current one but I'd be happy to merge a PR that uses a renderer option to change that behaviour.

lorisleiva:
Yes, consumers trying to export for both CJS and ESM will now need a build step to ensure this ESM syntax works with CJS exports. Again, fair play if you want to go ESM-only but this should be opt-in.

lorisleiva:
I will close this for now. Feel free to open a new PR that parameterise this change on the JS renderer. 🙏

PR #654, opened 2025-06-17
refactor(renderers-rust): extract solana-cpi, solana-msg and cleanup

sonicfromnewyoke:
### Problem

renderers-rust use bloated `solana-sdk` and `solana-program` crates

### Solution

extract `solana-cpi`, `solana-msg` from `solana-program` as a part of more granular migration, agreed in https://github.com/codama-idl/codama/pull/616

also cu up some loose ends and remove `solana-program` as a dependency

PR #685, opened 2025-07-05
refactor(renderers-vixen-parser): replace `solana-program` with more granular crates

sonicfromnewyoke:
### Problem

`renderers-vixen-parser` use bloated `solana-program` crate

### Solution

replace `solana-program` with `solana-*` crates to make deps more granular, agreed in #616 

cc @fernandodeluret

### DIFF
--- a/.changeset/eleven-radios-obey.md
+++ b/.changeset/eleven-radios-obey.md
@@ -1,0 +1,5 @@
+---
+'@codama/renderers-vixen-parser': minor
+---
+
+use granular solana-\* crates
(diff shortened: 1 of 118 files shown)


sonicfromnewyoke:
> I just had a 1st look at this, it's looking good. My only question is, if we update this in the vixen generator but not on the rust sdk generator will not cause cargo issues?

rust sdk was updated in the previous 4 MRs (#641, #645, #647, #654)

lorisleiva (approved):

lorisleiva:
Thank you! @fernandodeluret are you happy for me to publish a new version now (this will be a minor bump according to the changeset of this PR)?

PR #721, opened 2025-07-21
[renderers-rust] Export discriminator constants for accounts and instructions

lorisleiva:
This PR exports a constant for each `DiscriminatorNode` attached to an account or instruction that has a value.

For instance,

```rust
pub const TRANSFER_SOL_DISCRIMINATOR: u32 = 2;
```

or 

```rust
pub const MY_ACCOUNT_DISCRIMINATOR: [u8; 8] = [185, 149, 156, 78, 245, 108, 172, 68];
```

### DIFF
--- a/packages/renderers-rust/src/utils/discriminatorConstant.ts
+++ b/packages/renderers-rust/src/utils/discriminatorConstant.ts
@@ -1,0 +1,107 @@
+import {
+    camelCase,
+    ConstantDiscriminatorNode,
+    DiscriminatorNode,
+    FieldDiscriminatorNode,
+    InstructionArgumentNode,
+    isNode,
+    isNodeFilter,
+    snakeCase,
+    StructFieldTypeNode,
+    VALUE_NODES,
+} from '@codama/nodes';
+import { visit } from '@codama/visitors-core';
+
+import { getTypeManifestVisitor, TypeManifest } from '../getTypeManifestVisitor';
+import { ImportMap } from '../ImportMap';
+import { renderValueNode } from '../renderValueNodeVisitor';
+import { GetImportFromFunction } from './linkOverrides';
+
+type Fragment = { imports: ImportMap; render: string };
+
+function mergeFragments(fragments: Fragment[], merge: (parts: string[]) => string): Fragment {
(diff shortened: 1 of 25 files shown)


packages/renderers-rust/src/utils/discriminatorConstant.ts
```diff
@@ -0,0 +1,107 @@
...
+    const field = fields.find(f => f.name === discriminatorNode.name);
+    if (!field || !field.defaultValue || !isNode(field.defaultValue, VALUE_NODES)) {
+        return null;
+    }
+
+    const name = camelCase(`${prefix}_${discriminatorNode.name}`);
+    const typeManifest = visit(field.type, typeManifestVisitor);
+    const value = renderValueNode(field.defaultValue, getImportFrom);
+    return getConstant(name, typeManifest, value);
+}
```
buffalojoec, line 102: This is great for producing a constant, but does it provide the default value for the field itself on the struct? This is probably not easy to do, since you'd have to reason about a sensible Rust API:

```rust
struct MyStruct {
    discriminator: [u8; 4],
    message: String,
}

impl MyStruct {
    fn new(discriminator: [u8; 4], message: String) {
        // ...
    }

    fn new_with_defaults(message: String) {
        // ...
    }
}
```

We don't need to figure it out right now, but maybe an API like this could be useful for Codama? Even the JS renderer doesn't foray into this API work yet. The cool part is, you've already got a standard for struct fields by way of the `defaultValue` node, so you can just vacuum up all default nodes and use those to create the `new_with_defaults` constructor.

buffalojoec (approved):
Beautiful! Thanks for adding this.

packages/renderers-rust/src/utils/discriminatorConstant.ts
lorisleiva: The JS renderer actually does this because in JS land we have access to the Codec API so we can just have an `Encoder` function that automatically fills the discriminator for you (or any other field with default value) if it's not provided.

The main issue with the Rust renderer is that we don't have such high-level API for byte layouts so it makes the job of generating valid code much more complex.

My goal is to one day redesign this entire Rust renderer once we have a better foundation for it to build on if that makes sense.

PR #730, opened 2025-07-24
Tighten the return type of encode/decode functions for FixedSize cases

mcintyre94:
In #718 we use a FixedSize{Encoder/Decoder/Codec} to type functions such as `getXEncoder`.

In some cases (such as a discriminated union) this can generate a function that expects to return `FixedSize{Encoder}`, and in practice will do so, but is not guaranteed to do so by the Kit types.

Eg:
```ts
export function getBurnArgsEncoder(): FixedSizeEncoder<BurnArgsArgs> {
    return getDiscriminatedUnionEncoder([['V1', getStructEncoder([['amount', getU64Encoder()]])]]);
}
```

In Kit, `getDiscriminatedUnionEncoder` will return us a `FixedSizeEncoder` (by the same logic #718 determines it should) but does not guarantee this at a type level, as other combinations of variants can cause it not to do so. See https://github.com/anza-xyz/kit/blob/771f8aef1f8c096450c6e4ac05b8611150201485/packages/codecs-data-structures/src/union.ts#L136-L154

This is a Typescript error. This PR fixes the error by using `as` to cast the result of `getDiscriminatedUnionEncoder` to `FixedSizeEncoder<BurnArgsArgs>`. This is correct as long as the logic we have in both Codama and Kit to determine whether it's fixed size is correct. 

It uses the same logic as #718 to determine whether we need to do this - if we are using `FixedSize{Encoder}` then we cast to it. This covers cases where we don't need to use the cast, ie where the Kit codec does return a `FixedSize{Encoder}`, but it will just be a typescript no-op in those cases.

It doesn't cast to `unknown` so we still have a degree of type safety. I think the only alternative would be to create separate functions (or overrides) for `getUnion{Encoder}` in Kit, so that it returns `FixedSize{Encoder}` in these cases. I'm not sure how tricky this would be to type.


packages/renderers-js/e2e/anchor/idl.json
```diff
@@ -366,6 +366,15 @@
...
   "types": [
+    {
+      "name": "BurnArgs",
+      "type": {
+        "kind": "enum",
+        "variants": [
+          { "name": "V1", "fields": [{ "name": "amount", "type": "u64" }] }
+        ]
+      }
+    },
```
mcintyre94, line 377: This is from the `mpl-token-metadata` program and fails the test on `main` due to this bug

lorisleiva:
I wonder if we're not better off adding an option on the `getByteSizeVisitor` like `variableEnums: boolean` that defaults to `false` (current behaviour) but we can then toggle that to `true` on the JS renderer since we know Kit will always render a non-fixed codec?

PR #731, opened 2025-07-24
Fix size computation of array-like nodes

lorisleiva:
This PR fixes an issue with the `getByteSizeVisitor` that wrongly assessed array-like nodes such as `ArrayTypeNode`, and `SetTypeNode` and `MapTypeNode`.


mcintyre94 (approved):
LGTM!

PR #739, opened 2025-08-02
Update generated JS code to be more linter/formatter friendly

macalinao:
A notable change is using semicolons after error codes in constants.

This extracts the non-ESM specific logic in this previously closed PR: https://github.com/codama-idl/codama/pull/676


lorisleiva (approved):
This is much better, thank you so much!

PR #746, opened 2025-08-04
Use NodePath<Node> instead of Node[] for node selector function

### LINKED ISSUE #741: `bottomUpTransformerVisitor` passes a `Node[]` to `select`
This visitor passes a path of nodes to the select function(s):

https://github.com/codama-idl/codama/blob/3efa50acf784fef97fd561c4e1cc632e262bd86c/packages/visitors-core/src/bottomUpTransformerVisitor.ts#L25-L27

I'm also seeing this in practice: 
Image

This does match [the documentation for `NodeSelector`](https://github.com/codama-idl/codama/tree/main/packages/visitors-core#selecting-nodes), the type being used for the `select` function:

> A NodeSelectorFunction of type (path: NodePath) => boolean. In this case, the provided function is used to determine if the last node in the provided NodePath should be selected.

But it doesn't match the example in [the documentation for the visitor](https://github.com/codama-idl/codama/tree/main/packages/visitors-core#selecting-nodes):

```ts
    {
        // Delete all account nodes that start with "m".
        select: ['[accountNode]', node => 'name' in node && node.name.startsWith('m')],
        transform: () => null,
    },
```

This is a type error because `node` is an array of nodes, not a single one:

Image

I'm not sure what the preferred API for this visitor is, but we need to either:

- Keep it as taking a path, and update the documentation/example
- Or, change it to take the last node of the stack (`stack.peek()` I think) instead, so that the example in the documentation works 

FWIW Kinobi only passed a single node to select, as can be seen here:

https://github.com/metaplex-foundation/mpl-token-metadata/blob/23aee718e723578ee5df411f045184e0ac9a9e63/configs/kinobi.cjs#L649-L661

mcintyre94:
- Callers now get the type `NodePath<Node>` instead of `Node[]`, which enables functions like `getLastNodeFromPath`
- Also updated the example in the README which treated the input as a `Node`

Closes #741

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,23 +13,23 @@
 > [!NOTE]  
 > Wait, wasn't this project called Kinobi before? We've [renamed this project to **Codama**](https://github.com/codama-idl/codama/pull/234).
 
-Codama is a tool that describes any Solana program in a standardised format called a **Codama IDL**. 
+Codama is a tool that describes any Solana program in a standardised format called a **Codama IDL**.
 
 A Codama IDL can be used to:
 
- - Create clients for your programs in various languages/frameworks
- - Make CLIs
- - Provide additional information to explorers
+- Create clients for your programs in various languages/frameworks
+- Make CLIs
+- Provide additional information to explorers
 
 ![Codama header: A small double-sided mind-map with the Codama logo in the middle. On the left, we see the various ways to get a Codama IDL from your Solana programs such as "Anchor Program" and "Shank macros". On the right, we see the various utility tools that are offered for the IDL such as "Rendering client code" or "Rendering documentation".](https://github.com/user-attachments/assets/029af336-ea71-4e7f-9612-ef5bb187e3a0)
 
 ## I'm a busy Anchor developer. How do I use Codama?
 
-If you're using [Anchor](https://www.anchor-lang.com/), Codama can be used to create a TypeScript client that works with [Solana Kit](https://github.com/anza-xyz/kit). This combination replaces the traditional `@coral-xyz/anchor` and `@solana/web3.js` packages, and can be used in both TypeScript tests, and the browser. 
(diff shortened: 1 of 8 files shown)


packages/visitors-core/src/bottomUpTransformerVisitor.ts
```diff
@@ -22,7 +23,7 @@ export function bottomUpTransformerVisitor<TNodeKind extends NodeKind = NodeKind
     const transformerFunctions = transformers.map((transformer): BottomUpNodeTransformer => {
         if (typeof transformer === 'function') return transformer;
         return (node, stack) =>
-            getConjunctiveNodeSelectorFunction(transformer.select)(stack.getPath())
+            getConjunctiveNodeSelectorFunction(transformer.select)(stack.getPath() as NodePath<Node>)
```
lorisleiva, line 26: Another thing we could try here is pass `REGISTERED_NODE_KINDS` as argument to say: "get me a path whose last node may be from any node kind". That way, we are asserting the last node is not empty and we should get the appropriate return type back.

```suggestion
            getConjunctiveNodeSelectorFunction(transformer.select)(stack.getPath(REGISTERED_NODE_KINDS))
```

lorisleiva: Alternatively, we can use the `isFilledNodePath` helper like so.

```suggestion
            {
                const path = stack.getPath();
                if (!isFilledNodePath(path)) {
                    return false; // Or throw error as it should never happen.
                }
                return getConjunctiveNodeSelectorFunction(transformer.select)(path);
            }
```

lorisleiva (commented):
Thanks you! I'm happy with the cast but offered some alternatives. Let me know what you think of them. CI isn't green but looks like it's just some linting issue.

README.md
```diff
@@ -13,23 +13,23 @@
 > [!NOTE]  
 > Wait, wasn't this project called Kinobi before? We've [renamed this project to **Codama**](https://github.com/codama-idl/codama/pull/234).
 
-Codama is a tool that describes any Solana program in a standardised format called a **Codama IDL**. 
+Codama is a tool that describes any Solana program in a standardised format called a **Codama IDL**.
```
mcintyre94, line 16: These README changes are from `lint:fix` I think, shall I revert?

lorisleiva: No that's fine, it's better like that. ☺️

PR #778, opened 2025-08-19
Allow specifying docs in addPdasVisitor

macalinao:
- **Allow specifying docs on PDA nodes in the addPdasVisitor**
- **Add test for docs**
- **add changeset**


lorisleiva (approved):
Thanks!

PR #791, opened 2025-08-21
Remove redundant semicolon from generation of discriminator condition fragments

macalinao:
- **Remove redundant semicolon from generation of discriminator condition fragments**
- **Add changeset**

Although Prettier removes this if formatting is enabled, this ensures the render map visitor is generating valid code.


lorisleiva (commented):
Cool, is that the only place where we double the semi-colon? Could you remind me where the semi-colon in place inside the `ifTrue` variable?

PR #793, opened 2025-08-23
Add type annotations to discriminator constants

macalinao:
- **Add type annotations to discriminator constants**
- **Changeset**
- **Regenerate e2e**

This allows for use of the `--isolatedDeclarations` flag in TypeScript, which theoretically speeds up typechecking.

### DIFF
--- a/packages/renderers-js/src/fragments/discriminatorConstants.ts
+++ b/packages/renderers-js/src/fragments/discriminatorConstants.ts
@@ -60,9 +60,10 @@ export function getConstantDiscriminatorConstantFragment(
     const suffix = index <= 0 ? '' : `_${index + 1}`;
 
     const name = camelCase(`${prefix}_discriminator${suffix}`);
+    const strictType = visit(discriminatorNode.constant.type, typeManifestVisitor).strictType;
     const encoder = visit(discriminatorNode.constant.type, typeManifestVisitor).encoder;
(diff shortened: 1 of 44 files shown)


packages/renderers-js/src/fragments/discriminatorConstants.ts
```diff
@@ -81,27 +82,30 @@ export function getFieldDiscriminatorConstantFragment(
...
 
     return mergeFragments(
         [
-            mapFragmentContent(value, c => `export const ${constantName} = ${c};`),
+            mapFragmentContent(value, c => `export const ${constantName}: ${strictType.content} = ${c};`),
             mapFragmentContent(
                 encoder,
-                c => `export function ${constantFunction}() { return ${c}.encode(${constantName}); }`,
+                c =>
+                    `export function ${constantFunction}(): ReadonlyUint8Array { return ${c}.encode(${constantName}); }`,
```
lorisleiva, line 108: Whenever you introduce a new item that needs to be imported, it should be part of the fragment's import. Using the new Fragment API, you'll be able to do something like:

```ts
const readonlyUint8Array = use("ReadonlyUint8Array", "solanaCodecsCore");
return fragment`export function ${constantFunction}():${readonlyUint8Array} { return ${encoder}.encode(${constantName}); }`
```

lorisleiva (changes requested):
Just a small comment but I think this one could do with a rebase too. Sorry for the late review.

PR #799, opened 2025-08-27
Add support for PDA derivation and additional programs in the Rust renderer

macalinao:
Note: I added Quarry's Codama IDL here since none of the test cases had PDAs defined. There is the token program, but we cannot serialize `Option` in Borsh.


lorisleiva (changes requested):
Would you mind rebasing and then I'll ask Febo to give this a look since he's more used to the Rust renderer than I am?

PR #840, opened 2025-09-10
Add `getMaxByteSizeVisitor` helper

lorisleiva:
This PR adds a new `getMaxByteSizeVisitor` helper that returns the maximum size in bytes a node can take, or `null` when no max size can be inferred.

This is slightly different to the existing `getByteSizeVisitor` that returns a number if and only if the node is of fixed size.

For instance, take the following node: `optionTypeNode(numberTypeNode("shortU16"))`.

Here the byte size visitor would return `null` because the `shortU16` number can take between 1 to 3 bytes and the option wrapper can either be of size `1` if the value is `None` or `size(x) + 1` if the value is `Some(x)`.

However, the max size visitor can understand that this node will take at most 4 bytes: 1 for the option prefix and 3 for the maximum value of a `shortU16`.

### DIFF
--- a/.changeset/purple-sites-behave.md
+++ b/.changeset/purple-sites-behave.md
@@ -1,0 +1,5 @@
+---
+'@codama/visitors-core': patch
+---
+
+Add `getMaxByteSizeVisitor` helper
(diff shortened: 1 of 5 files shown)


febo (approved):
Looks great!

Issue #263, opened 2024-10-29
Add instruction status such as deprecated or archived


lorisleiva:
It is not uncommon for program developers to sunset their instructions by first, logging a deprecation warning and then removing it completely by throwing an error when the archived instruction is recognised by the program.

It would be nice for Codama to keep track of that information via a status attribute or even a dedicated `InstructionStatusNode`. The instruction status could be things like:
- `live` if the instruction is accessible (the default state).
- `deprecated` if the instruction is about to be archived.
- `archived` if the instruction is no longer accessible. Note that this is better than simply removing the instruction from the Codama IDL as explorers would still need to parse old instruction for the program.

We could even image more advanced statuses like:
- `draft` if the instruction is accessible but not fully implemented yet.
- `unaudited` if the instruction is accessible and fully implemented but no audit was performed for it yet.

Additionally, an optional `statusMessage` attribute could be useful for program maintainers to provide more information about the current status to their consumers.

lorisleiva:
Implemented by @TanmayDhobale. 
- [JavaScript PR](https://github.com/codama-idl/codama/pull/936).
- [Rust PR](https://github.com/codama-idl/codama-rs/pull/75).

Thank you! 🍺

PR #936, opened 2025-11-27
feat: add status and statusMessage to InstructionNode

tanmay4l:
#263  cc @lorisleiva

### DIFF
--- a/.changeset/wet-meals-work.md
+++ b/.changeset/wet-meals-work.md
@@ -4,4 +4,4 @@
 '@codama/nodes': minor
 ---
 
-Rename InstructionStatusNode field from status to lifecycle and simplify API signature
+Add a new InstructionStatusNode to instructions
--- a/README.md
+++ b/README.md
@@ -114,15 +114,14 @@ Feel free to PR your own visitor here for others to discover. Note that they are
 
 ### Generates program clients
 
-| Visitor                                                                                         | Description                                                                                             | Maintainer                            |
-| ----------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------- | ------------------------------------- |
-| `@codama/renderers-js` ([docs](https://github.com/codama-idl/renderers-js))                     | Generates a JavaScript client compatible with [Solana Kit](https://www.solanakit.com/).                 | [Anza](https://www.anza.xyz/)         |
-| `@codama/renderers-js-umi` ([docs](https://github.com/codama-idl/renderers-js-umi))             | Generates a JavaScript client compatible with [the Umi framework](https://developers.metaplex.com/umi). | [Metaplex](https://www.metaplex.com/) |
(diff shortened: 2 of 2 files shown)


TanmayDhobale:
> Hey, thanks for tackling this! How would you feel about creating a new `InstructionStatusNode` instead that attaches to the `InstructionNode` in an optional `status` attribute?
> 
> Then we can have a helper in `@codama/nodes` like `status: InstructionStatusNode("deprecated", "Some deprecated message")`.
> 
> We'll also need to update core visitors so that the new node is registered properly and release a matching PR in `codama-rs` repo so that the standard is kept in-sync in both languages.
> 
> Finally, I think we can drop the `"unaudited"` status for now. This was a passing thought I had and thinking about it more I think the audit status should probably be more at the program level than the instruction level (something we can work on some other time).

yep sounds good to me so status can be transformed/validated separately

TanmayDhobale:
hey @lorisleiva created InstructionStatusNode as a separate node attached to InstructionNode as optional status  added the helper in @codama/nodes and updated all core visitors. also got the matching codama-rs pr ready will raised later 

 one que  changeset bot flagged missing changeset do i need to add one or is that handled separately?

packages/nodes/src/InstructionStatusNode.ts
```diff
@@ -0,0 +1,14 @@
+import type { InstructionStatusNode, InstructionStatus } from '@codama/node-types';
+
+export function instructionStatusNode(
+    status: InstructionStatus,
+    options: { message?: string } = {},
+): InstructionStatusNode {
```
lorisleiva: I think it'd be nicer to use this helper like this: `instructionStatusNode("deprecated", "Use this other instruction instead")`.

```suggestion
export function instructionStatusNode(
    status: InstructionStatus,
    message?: string,
): InstructionStatusNode {
```

What do you think?

packages/node-types/src/InstructionStatusNode.ts
```diff
@@ -0,0 +1,9 @@
+import type { InstructionStatus } from './shared';
+
+export interface InstructionStatusNode {
+    readonly kind: 'instructionStatusNode';
+
+    // Data.
+    readonly status: InstructionStatus;
```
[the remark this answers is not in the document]
TanmayDhobale: yea works, instruction.status.status is terrible ux

packages/nodes/src/InstructionStatusNode.ts
TanmayDhobale: yea make sense , thanks

lorisleiva (commented):
Tysm! Just a couple of small comments and we're there.

EDIT: One of the tests is failing as well I think because of a `status` residue somewhere.

TanmayDhobale:
> For the lint issues, you should be able to run `pnpm lint:fix` on the root directory.

Linting should be fixed now :)

PR #953, opened 2025-12-11
fix: impl nested account name deduplication

### LINKED ISSUE #754: [nodes-from-anchor] Issue with renderers-js and nested account names
If one has an Anchor program with two separate nested accounts with the same account name, Codama generates TypeScript code with duplicate property names.

Below is my fix for this issue-- is there a recommended place for this to go in the codebase?

```typescript
import type {
  AnchorIdl,
  IdlV01InstructionAccountItem,
} from "@codama/nodes-from-anchor";
import { instructionAccountNodeFromAnchorV01 } from "@codama/nodes-from-anchor";
import type {
  AccountNode,
  InstructionAccountNode,
  InstructionArgumentNode,
} from "codama";
import {
  assertIsNode,
  bottomUpTransformerVisitor,
  camelCase,
  rootNodeVisitor,
  visit,
} from "codama";

function instructionAccountNodesFromAnchorV01(
  allAccounts: AccountNode[],
  instructionArguments: InstructionArgumentNode[],
  idl: IdlV01InstructionAccountItem[],
  parent: string | null = null,
): InstructionAccountNode[] {
  return idl.flatMap((account) =>
    "accounts" in account
      ? instructionAccountNodesFromAnchorV01(
          allAccounts,
          instructionArguments,
          account.accounts,
          // null,
          parent ? `${parent}_${account.name}` : account.name,
        )
      : [
          instructionAccountNodeFromAnchorV01(
            allAccounts,
            instructionArguments,
            {
              ...account,
              name: parent ? `${parent}_${account.name}` : account.name,
              pda:
                account.pda && parent
                  ? {
                      ...account.pda,
                      seeds: account.pda.seeds.map((seed) => {
                        if (seed.kind === "account") {
                          return {
                            ...seed,
                            path: `${parent}_${seed.path}`,
                          };
                        }
                        return seed;
                      }),
                    }
                  : undefined,
            },
            idl,
          ),
        ],
  );
}

export function instructionAccountsDedupeVisitor(idl: AnchorIdl) {
  return rootNodeVisitor((node) => {
    const accountNodes = node.program.accounts;
    const instructionVisitor = bottomUpTransformerVisitor([
      {
        select: "[instructionNode]",
        transform: (node, _stack) => {
          assertIsNode(node, "instructionNode");
          const idlIx = idl.instructions.find(
            (ix) => camelCase(ix.name) === node.name,
          );
          if (!idlIx) {
            throw new Error(`Instruction ${node.name} not found in Idl`);
          }
          return {
            ...node,
            accounts: instructionAccountNodesFromAnchorV01(
              accountNodes,
              node.arguments,
              idlIx.accounts,
            ),
          };
        },
      },
    ]);
    return visit(node, instructionVisitor);
  });
}
```

AvhiMaz:
Automatically prefix nested account names with their parent group name 
when duplicates are detected, preventing invalid TypeScript output.
Only applies prefixing when necessary, preserving original names otherwise.

Fixes #754

### DIFF
--- a/packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
+++ b/packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
@@ -14,7 +14,11 @@ import {
 } from '@codama/nodes';
 import { expect, test } from 'vitest';
 
-import { instructionAccountNodeFromAnchorV01, instructionAccountNodesFromAnchorV01 } from '../../src';
+import {
+    IdlV01InstructionAccountItem,
+    instructionAccountNodeFromAnchorV01,
+    instructionAccountNodesFromAnchorV01,
+} from '../../src';
 
 test('it creates instruction account nodes', () => {
     const node = instructionAccountNodeFromAnchorV01(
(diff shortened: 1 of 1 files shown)


packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
```diff
@@ -17,26 +18,58 @@ import {
...
+    return checkDuplicates(idl);
+}
+
 export function instructionAccountNodesFromAnchorV01(
     idl: IdlV01InstructionAccountItem[],
     instructionArguments: InstructionArgumentNode[],
+    prefix?: string,
 ): InstructionAccountNode[] {
+    const hasDuplicates = hasDuplicateAccountNames(idl);
+    const shouldPrefix = prefix !== undefined || hasDuplicates;
```
lorisleiva: I think this should be like the v00 implementation to avoid checking for duplicates if we already have a prefix.


```suggestion
    const shouldPrefix = prefix !== undefined || hasDuplicateAccountNames(idl);
```

packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
```diff
@@ -115,6 +115,35 @@ test('it flattens nested instruction accounts', () => {
     ]);
 });
 
+test('it prevents duplicate names by prefixing nested accounts with different parent names', () => {
```
lorisleiva: Thanks for adding tests! Would you mind adding another one that checks at least two layers of nested accounts here as well?

Issue #69, opened 2024-06-18
[nodes] Add `ConstantNodes`


lorisleiva:
Add support for `constants` array with a new `ConstantNode` inside `ProgramNode`.

Then update the `nodes-from-anchor` package to include them.

Then update the renderers to render the constants.

PR #958, opened 2025-12-12
feat(nodes): add ConstantNode support with Anchor parsing

tanmay4l:
#69  cc @lorisleiva

### DIFF
--- a/packages/node-types/src/ConstantNode.ts
+++ b/packages/node-types/src/ConstantNode.ts
@@ -1,0 +1,15 @@
+import type { CamelCaseString, Docs } from './shared';
+import type { TypeNode } from './typeNodes/TypeNode';
+import type { ValueNode } from './valueNodes/ValueNode';
+
+export interface ConstantNode<TType extends TypeNode = TypeNode, TValue extends ValueNode = ValueNode> {
+    readonly kind: 'constantNode';
+
+    // Data.
+    readonly name: CamelCaseString;
+    readonly docs?: Docs;
+
+    // Children.
+    readonly type: TType;
+    readonly value: TValue;
+}
--- a/packages/nodes-from-anchor/src/utils.ts
+++ b/packages/nodes-from-anchor/src/utils.ts
@@ -1,3 +1,27 @@
+import { bytesValueNode, numberValueNode, stringValueNode, ValueNode } from '@codama/nodes';
+
 export function hex(bytes: number[] | Uint8Array): string {
-    return (bytes as number[]).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
+    return Array.from(bytes).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
+}
+
+export function parseConstantValue(valueString: string): ValueNode {
+    if (valueString.startsWith('[') && valueString.endsWith(']')) {
+        // It's a byte array
+        try {
+            const bytes = JSON.parse(valueString) as number[];
+            const uint8Array = new Uint8Array(bytes);
+            return bytesValueNode('base16', hex(uint8Array));
+        } catch {
+            // Fallback to string if parsing fails
+            return stringValueNode(valueString);
+        }
+    }
+
+    if (/^-?\d+$/.test(valueString)) {
+        // It's a number
+        return numberValueNode(Number(valueString));
+    }
+
+    // It's a string
+    return stringValueNode(valueString);
 }
--- a/packages/nodes-from-anchor/src/utils.ts
+++ b/packages/nodes-from-anchor/src/utils.ts
@@ -1,27 +1,45 @@
-import { bytesValueNode, numberValueNode, stringValueNode, ValueNode } from '@codama/nodes';
+import {
+    booleanValueNode,
+    bytesValueNode,
+    isNode,
+    numberValueNode,
+    publicKeyValueNode,
+    stringTypeNode,
+    stringValueNode,
+    TypeNode,
+    ValueNode,
+} from '@codama/nodes';
 
 export function hex(bytes: number[] | Uint8Array): string {
     return Array.from(bytes).reduce((str, byte) => str + byte.toString(16).padStart(2, '0'), '');
 }
 
-export function parseConstantValue(valueString: string): ValueNode {
-    if (valueString.startsWith('[') && valueString.endsWith(']')) {
-        // It's a byte array
+export function parseConstantValue(valueString: string, type: TypeNode): { type: TypeNode; value: ValueNode } {
+    if (isNode(type, 'bytesTypeNode')) {
         try {
             const bytes = JSON.parse(valueString) as number[];
-            const uint8Array = new Uint8Array(bytes);
-            return bytesValueNode('base16', hex(uint8Array));
+            return { type, value: bytesValueNode('base16', hex(new Uint8Array(bytes))) };
         } catch {
-            // Fallback to string if parsing fails
-            return stringValueNode(valueString);
+            return { type: stringTypeNode('utf8'), value: stringValueNode(valueString) };
         }
     }
 
-    if (/^-?\d+$/.test(valueString)) {
-        // It's a number
-        return numberValueNode(Number(valueString));
+    if (isNode(type, 'numberTypeNode')) {
+        if (/^-?\d+$/.test(valueString)) {
+            return { type, value: numberValueNode(Number(valueString)) };
+        }
+        return { type: stringTypeNode('utf8'), value: stringValueNode(valueString) };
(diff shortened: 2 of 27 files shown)


lorisleiva:
Hi, thanks for the PR!

This is looking pretty good. The only thing I'm unsure of is whether on not the `ConstantNode` should wrap the existing [`ConstantValueNode`](https://github.com/codama-idl/codama/blob/main/packages/nodes/docs/valueNodes/ConstantValueNode.md) which is the existing way for us to describe a `type + value` combo.

On one hand, this would make the `ConstantValueNode` the single source of truth for handling constants (visitors just need to add an implementation for this Node and it'll work everywhere in the tree).

On the other hand, this make the `ConstantNode` a bit awkward because it'll access data by doing something like `program.constants[0].constant.value`. Maybe something that can be fixed with a better naming convention?

What do you think?

packages/nodes/docs/ConstantNode.md
```diff
@@ -0,0 +1,95 @@
...
+```ts
+const seedPrefix = constantNode('SEED_PREFIX', bytesTypeNode(), bytesValueNode('base16', '74657374'));
+```
+
+### With Documentation
+
+```ts
+const maxItems = constantNode('MAX_ITEMS', numberTypeNode('u64'), numberValueNode(1000));
+// Note: Add docs via the ConstantNode interface if needed
+```
```
lorisleiva: This example needs to use the 4th parameter to show how to document a constant.

lorisleiva (changes requested):
Thanks for this PR and again sorry for the super later review.

I've made a few comments, lmk what you think.

I also think you're **missing an important test** file in the `packages/visitors-core/test/nodes` folder (e.g. look at the `packages/visitors-core/test/nodes/DefinedTypeNode.test.ts` file for reference).

P.S.: Don't forget to rebase. ☺️

alexanderguy:
Bump for this feature getting in.  We'd like to use it in https://github.com/faremeter/flex .  We can help with review/getting it landed, if anyone needs anything.

tanmay4l:
> Bump for this feature getting in. We'd like to use it in https://github.com/faremeter/flex . We can help with review/getting it landed, if anyone needs anything.

I apologize for the delay. I’m back now and will push the updated code soon.

packages/node-types/src/ConstantNode.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface ConstantNode<TType extends TypeNode = TypeNode, TValue extends ValueNode = ValueNode> {
+    readonly kind: 'constantNode';
+
+    // Data.
+    readonly name: CamelCaseString;
+    readonly docs?: Docs;
+
+    // Children.
+    readonly type: TType;
+    readonly value: TValue;
```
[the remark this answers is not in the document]
tanmay4l: sounds good, left it as-is then

packages/nodes-from-anchor/src/v00/ConstantNode.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function constantNodeFromAnchorV00(idl: Partial<IdlV00Constant>): ConstantNode {
+    const name = idl.name ?? '';
+    const valueString = idl.value ?? '';
+
+    // For constants, 'bytes' type represents a raw byte array, not a sized string
+    // so we use u8 to represent the type of each element
+    const type =
+        idl.type === 'bytes'
+            ? numberTypeNode('u8')
```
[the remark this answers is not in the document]
tanmay4l: switched to bytesTypeNode().

packages/nodes-from-anchor/test/v00/ConstantNode.test.ts
```diff
@@ -0,0 +1,73 @@
...
+
+test('it handles malformed JSON in value gracefully', () => {
+    const node = constantNodeFromAnchorV00({
+        name: 'invalid_bytes',
+        type: 'bytes',
+        value: '[invalid json',
+    });
+
+    // Should fallback to string value
+    expect(node.value).toEqual(stringValueNode('[invalid json'));
```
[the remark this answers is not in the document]
tanmay4l: done, full toEqual now

packages/nodes-from-anchor/test/v01/ConstantNode.test.ts
```diff
@@ -0,0 +1,83 @@
...
+});
+
+test('it parses constant with bytes type and value', () => {
+    const node = constantNodeFromAnchorV01({
+        name: 'seed_prefix',
+        type: 'bytes',
+        value: '[116, 101, 115, 116]', // "test" in bytes
+    });
+
+    expect(node).toEqual(constantNode('seedPrefix', numberTypeNode('u8'), bytesValueNode('base16', '74657374')));
```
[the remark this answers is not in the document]
tanmay4l: my bad , applied

packages/nodes-from-anchor/src/v01/ConstantNode.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function constantNodeFromAnchorV01(idl: Partial<IdlV01Const>): ConstantNode {
+    const name = idl.name ?? '';
+    const valueString = idl.value ?? '';
+
+    const declaredType =
+        idl.type === 'bytes'
+            ? bytesTypeNode()
+            : idl.type
+              ? typeNodeFromAnchorV01(idl.type, { constArgs: {}, typeArgs: {}, types: {} })
```
lorisleiva: You could pass `generics` from the `programNodeFromAnchorV01` function here.

packages/nodes-from-anchor/src/v01/ProgramNode.ts
```diff
@@ -24,6 +25,7 @@ export function programNodeFromAnchorV01(idl: IdlV01): ProgramNode {
 
     return programNode({
         accounts: accountNodes,
+        constants: (idl.constants ?? []).map(constantNodeFromAnchorV01),
```
lorisleiva: nit: maybe follow the same convention as the other program items by having `const constants = idl.constants ?? [];` defined above.

As per my previous comment, this can then become:

```ts
    constants: constants.map(constant => constantNodeFromAnchorV01(constant, generics)),
```

packages/nodes-from-anchor/src/utils.ts
```diff
@@ -1,3 +1,52 @@
...
+        } catch {
+            // fall through to the string fallback below
+        }
+        return { type: stringTypeNode('utf8'), value: stringValueNode(valueString) };
+    }
+
+    if (isNode(type, 'numberTypeNode')) {
+        if (/^-?\d+$/.test(valueString)) {
+            return { type, value: numberValueNode(Number(valueString)) };
+        }
```
lorisleiva: Do we need to worry about decimal numbers here?

packages/nodes-from-anchor/test/v00/ConstantNode.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    numberTypeNode,
+    numberValueNode,
+    stringTypeNode,
+    stringValueNode,
+} from '@codama/nodes';
+import { expect, test } from 'vitest';
+
+import { constantNodeFromAnchorV00, programNodeFromAnchorV00 } from '../../src';
+
+test('it parses constant with number type and value', () => {
```
lorisleiva, line 19: nit: could we add test for public keys and booleans too? Same for v01 tests.

packages/nodes-from-anchor/test/v00/ConstantNode.test.ts
```diff
@@ -0,0 +1,95 @@
...
+    const node = constantNodeFromAnchorV00({
+        name: 'seed_prefix',
+        type: 'bytes',
+        value: '[116, 101, 115, 116]', // "test" in bytes
+    });
+
+    expect(node).toEqual(constantNode('seedPrefix', bytesTypeNode(), bytesValueNode('base16', '74657374')));
+});
+
+test('it parses constant with string value', () => {
```
lorisleiva: nit: the test title is a bit misleading. We're most testing that "linked" defined type are resolved as strings (instead of actually following that type's definition). Same comment for the matching V01 test.

lorisleiva (changes requested):
Thank for the changes! I've just got a few minor comments left but after that we should be good to merge.

lorisleiva (approved):
Thank you! Do you think you'd be able to reproduce that change on the `codama-rs` side? 🙏

tanmay4l:
> Thank you! Do you think you'd be able to reproduce that change on the `codama-rs` side? 🙏

on it ser

PR #980, opened 2026-02-08
doc: add `renderers-go` to the list of program clients

sonicfromnewyoke:
### Problem

**Generates program clients** lacks of reference to the `renderers-go` package

### Solution

add missed links which lead to the Go SDK and renderer itself


lorisleiva (approved):
Thank you!

PR #987, opened 2026-03-25
feat: add dynamic-client

mikhd:
## Summary

Adds `@codama/dynamic-client` runtime instruction builder that creates `Instruction` from Codama IDLs without code generation.

## Key Features

- **Anchor-like builder API** — `programClient.methods.transfer({}).accounts({}).instruction()`.
- **Auto-resolution** — Accounts with `defaultValue` (PDAs, program IDs, constants) are resolved automatically.
- **Typed & untyped** — Use without or with types `createProgramClient<MyProgramClient>(idl)`.
- **Standalone PDA derivation** — `programClient.pdas.myPda({seeds})`.
- **CLI** — Generate `ProgramClient` TypeScript types with `npx @codama/dynamic-instructions generate-client-types <idl.json> <output-dir>` for a given program.

## Package Structure
```
  packages/dynamic-client/
  ├── src/
  │   ├── index.ts                      # Public API entry point
  │   ├── types/                        # Type re-exports
  │   ├── program-client/               # Public API client
  │   │   ├── create-program-client.ts  #   Client creation entry point
  │   │   ├── methods-builder.ts        #   Client methods builder for creating Instruction
  │   │   ├── collect-pdas.ts           #   PDA collection from IDL
  │   │   └── derive-standalone-pda.ts  #   Standalone PDA derivation
  │   ├── instruction-encoding/         # Instruction creation
  │   │   ├── instructions.ts           #   Instruction encoding pipeline
  │   │   ├── validators.ts             #   Validators for accounts, arguments and nodes (superstruct)
  │   │   ├── accounts/                 #   AccountMeta creation and validation
  │   │   ├── arguments/                #   Arguments encoding and validation
  │   │   ├── resolvers/                #   Account/PDA/conditional resolution
  │   │   └── visitors/                 #   IDL tree-traversal visitors
  │   ├── shared/                       # Utilities
  │   └── cli/                          # CLI with type generation (commanderjs)
  │       └── commands/
  ├── test/
  │   ├── unit/                         # Unit tests
  │   ├── programs/                     # Program integration tests via LiteSVM
  │   │   ├── system-program/           #   System program
  │   │   ├── token/                    #   SPL Token program
  │   │   ├── token-2022/               #   Token-2022 program
  │   │   ├── associated-token-account/ #   ATA program
  │   │   ├── mpl-token-metadata/       #   MPL Token Metadata program
  │   │   ├── pmp/                      #   PMP program
  │   │   ├── sas/                      #   SAS program
  │   │   ├── anchor/                   #   Custom Anchor program (custom and edge cases)
  │   │   ├── circular-account-refs/    #   Circular dependency edge case
  │   │   ├── custom-resolvers/         #   Custom resolver integration
  │   │   ├── idls/                     #   Codama IDL JSONs
  │   │   ├── generated/                #   Auto-generated types from Codama idls/
  │   │   └── dumps/                    #   Program .so dumps for LiteSVM
  │   └── svm-test-context.ts           # LiteSVM wrapper
```

## Instruction Building Pipeline

```
const client = createProgramClient<MyProgramClientType>(idl) // ProgramClient

client.methods.<instructionName>(arguments)
    .accounts({}) // required accounts
    .signers([]) // clarified "either" signers
    .resolvers({}) // resolvers for ResolverValueNode
    .instruction() // Instruction
```

#### Under the hood it:

1. Validates arguments and accounts input.
2. Resolves arguments from custom resolvers (optionally).
3. Encodes arguments to bytes.
4. Resolves account addresses (PDAs, program IDs, conditionals).
5. Returns final `Instruction { programAddress, accounts, data }`.

## Tests

- **test/unit** — contains tests for the package.
- **test/program** — contains tests of ProgramClient public interface in LiteSVM environment. ProgramClient is used for building and sending instructions to Solana programs.

### Notes:

- CI/CD update:
  - We're using Anchor@0.32.1 for simulation of some custom IDL cases, thus Anchor and Rust are required during tests.
  - Solana version was updated v2 to v3 due to [issues](https://solana.stackexchange.com/questions/24011/anchor-build-fails-due-to-constant-time-eq-v0-4-2-requiring-rust-edition2024-ca) with anchor spl-token compilation.


lorisleiva (commented):
Hey guys, thank you so much for the heroic amount of work here. It's very hard for me to review everything thoroughly so I thought I'd try a quick "first pass" at the PR but randomly jumping into various source files and giving my first impressions. I apologise in advance if I lacked context within my comments but hopefully you can rectify me pretty quickly if that's the case. There's a mixture of small comments and comments that could lead to significant changes so please let me know if you don't have enough time to dedicate to some of these.

packages/dynamic-client/src/instruction-encoding/resolvers/resolve-conditional.ts
```diff
@@ -0,0 +1,71 @@
...
+export type ResolveConditionalContext = BaseResolutionContext & {
+    conditionalValueNode: ConditionalValueNode;
+    ixAccountNode: InstructionAccountNode;
+};
+
+/**
+ * Evaluates a ConditionalValueNode's condition.
+ * Returns the matching branch (ifTrue or ifFalse) as an InstructionInputValueNode or undefined if no branch matches.
+ */
+export async function resolveConditionalValueNodeCondition({
```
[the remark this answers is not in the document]
mikhd: Thank you! Can you please explain what was your thougs about its usage?

Not sure I understood it correctly, but we created a minimal example of `getResolvedInstructionInputsVisitor` integration in a separate PR https://github.com/hoodieshq/codama/pull/14.
[Here](https://github.com/hoodieshq/codama/pull/14/changes#diff-48d69d1bcc922fb2928748ad05b1fdf07ef072b2c8a1505e0c4fce2eac0eb116R108) promise.all accounts were replaced by sequential resolution from `getResolvedInstructionInputsVisitor`.

It can help with catching circular dependencies early, but other that that we still need visitors and resolvers to evaluate conditionals, derive pdas, encode seeds based on user inputs at runtime. Maybe we are addressing different things?

Also one real-world limitation we hit was that the `getResolvedInstructionInputsVisitor` visitor static analysis cant resolve nested argument dependencies in `resolverValueNode` with `dependsOn`.

For example, `create` instruction from Metaplex `mpl-token-metadata` has [resolverValueNode({ dependsOn: [argumentValueNode('tokenStandard')] })](https://github.com/hoodieshq/codama/blob/main-dynamic-instructions/packages/dynamic-instructions/test/programs/idls/mpl-token-metadata-idl.json#L5613), where `tokenStandard` is a `key` of a `struct createArgs {}`.

`getResolvedInstructionInputsVisitor` uses `getAllInstructionArguments` which only returns top-level args `discriminator, createArgs`, doesn't find "tokenStandard" and throws `CODAMA_ERROR__VISITORS__INVALID_INSTRUCTION_DEFAULT_VALUE_DEPENDENCY`.

`CodamaError: Dependency [tokenStandard] of kind [argumentValueNode] is not a valid dependency of [splTokenProgram] of kind [instructionAccountNode] in the [create] instruction`

Our solution was to use fallback strategy: when `getResolvedInstructionInputsVisitor` throws, we fallback to a custom iterative loop.

packages/dynamic-client/src/instruction-encoding/accounts/create-account-meta.ts
```diff
@@ -0,0 +1,142 @@
...
+ * Throws errors if required accounts are missing or cannot be resolved.
+ */
+export async function createAccountMeta(
+    root: RootNode,
+    ixNode: InstructionNode,
+    argumentsInput: ArgumentsInput = {},
+    accountsInput: AccountsInput = {},
+    signers: EitherSigners = [],
+    resolversInput: ResolversInput = {},
+): Promise<AccountMeta[]> {
```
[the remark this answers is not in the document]
mikhd: Thank you!
We checked NodeStack + LinkableDictionary adoption for the resolution pipeline.

Key constraint: all visitor infrastructure is synchronous (recordNodeStackVisitor does sync push/pop), while our resolution pipeline is async due to getProgramDerivedAddress and async custom resolver functions.

But [here](https://github.com/hoodieshq/codama/pull/15/) is a minimal example of what can be implemented (as I understood it). I guess, it doesn't fully align to your suggestion, but:
- The NodeStack holds the [Root, Program, Instruction] path. It is created once before the sequential resolution loop and never modified during async operations. NodeStack path replaces root/ixNode parameters with getting them from Stack.
- LinkableDictionary is created and passed as args for PDA and linked accounts lookup.

The main issue for full visitor conversion and push/pop NodeStack is that the current runtime resolution is not a normal tree transform - it is async. It also stateful and depends on runtime inputs: accountsInput, argumentsInput, custom resolvers.

What was your thoughts?

packages/dynamic-instructions/src/shared/errors.ts
```diff
@@ -0,0 +1,34 @@
+export class DynamicInstructionsError extends Error {
```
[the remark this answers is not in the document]
mikhd: Updated: https://github.com/codama-idl/codama/pull/987/commits/6549faaef8fe979564ff83e3019fd571f1ffda06

packages/dynamic-client/src/program-client/create-program-client.ts
```diff
@@ -0,0 +1,134 @@
...
+            },
+            has(target, prop) {
+                return Reflect.has(target, prop) || (typeof prop === 'string' && instructions.has(prop));
+            },
+        },
+    ) as ProgramClient['methods'];
+
+    const pdaNodes = collectPdaNodes(root);
+
+    const pdas =
```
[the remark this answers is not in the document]
lorisleiva: @mikhd Before I dig too much into the other parts of this review, I'd love your opinion on that one. I think splitting this package into 3 might make more sense and be consistent with the way we currently do things with `dynamic-codecs` and `dynamic-parsers`. Wdyt?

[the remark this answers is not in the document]
mikhd: Hey,
Reviewed this with guys, at the moment, it seem that the current package contains single united feature.
The question is would dynamic-instructions and dynamic-pdas be used as standalone packages or are they for internal-only usage?
It seems that renaming to dynamic-client makes sense but splitting into three packages would turn internal coupling into cross-package dependencies - more maintenance overhead without real decoupling.

lorisleiva:
@trevor-cortex Please provide a thorough review of this PR. Take all the time you need as this is a big PR and we wanna make sure we're not missing anything important before merging. Double check your findings before adding comments to your review to minimise noise. God speed.

packages/dynamic-client/src/cli/commands/generate-client-types/generate-client-types.ts
```diff
@@ -0,0 +1,333 @@
...
+
+    const pdaMap = collectPdaNodesFromIdl(idl);
+
+    const hasPdas = pdaMap.size > 0;
+    const addressImports = hasPdas ? 'Address, ProgramDerivedAddress' : 'Address';
+
+    let output = `/**
+ * Auto-generated types for strongly-typed ProgramClient
+ * DO NOT EDIT - Generated by @hoodieshq/dynamic-instructions generate-client-types
+ *
```
trevor-cortex, line 28: This should be `@codama/dynamic-instructions`, not `@hoodieshq/dynamic-instructions`.

```suggestion
 * DO NOT EDIT - Generated by @codama/dynamic-instructions generate-client-types
```

.github/actions/setup-anchor/action.yml
```diff
@@ -0,0 +1,63 @@
...
+inputs:
+  anchor-version:
+    description: Anchor CLI version to install (e.g. 0.32.1)
+    required: true
+  rust-version:
+    description: Rust toolchain version to install (e.g. 1.93.0)
+    required: true
+  cargo-lock-path:
+    description: Path to Cargo.lock for cache key hashing
+    required: false
```
trevor-cortex, line 13: These default paths reference `test/anchor/` but the actual directory is `test/programs/anchor/`. This will cause cache misses in CI (the `hashFiles` will always return `'no-lock'`).

```suggestion
    default: packages/dynamic-instructions/test/programs/anchor/Cargo.lock
```

trevor-cortex (commented):
## Summary

This is a substantial and well-structured PR that adds `@codama/dynamic-instructions` — a runtime instruction builder that creates `Instruction` objects from Codama IDLs without code generation. The package provides an Anchor-like builder API with auto-resolution of PDA accounts, program IDs, and conditional defaults.

The architecture is solid: clean separation between the public API (`program-client/`), instruction encoding pipeline (`instruction-encoding/`), and CLI. The visitor pattern for node resolution is well-applied. The test suite is impressively thorough — integration tests against real Solana programs (System, Token, ATA, MPL Token Metadata, PMP) via LiteSVM, plus edge cases for circular dependencies, custom resolvers, and collection types.

## Key Issues

### Must fix
1. **Wrong package name in generated types header** — `@hoodieshq/dynamic-instructions` should be `@codama/dynamic-instructions`
2. **CI path mismatch** — `setup-anchor/action.yml` defaults reference `test/anchor/` but the actual path is `test/programs/anchor/`
3. **Unused `Codec` value import** — should be `import type { Codec }` (tree-shaking concern)

### Worth reviewing
4. **Parallel account resolution with `Promise.all`** — resolving accounts in parallel may cause issues with dependent accounts that have shared mutable state (the `resolutionPath` circular dependency detection works correctly since each path is forked, but this is worth noting)
5. **`createAccountMeta` role downgrade** — all accounts matching `programAddress` get downgraded to readonly, but this only matters for optional accounts resolved to programId strategy. The check is broader than necessary.

## Areas for subsequent reviewers

- The `validators.ts` file (514 lines) is large — consider splitting into per-type-node validators in a future refactor
- The `generate-client-types.ts` string-concatenation approach works but could benefit from a proper AST builder if the generated output grows
- The `mapTypeNode` in `codamaTypeToTS` always emits `Record<string, V>` regardless of the actual key type — this is a known simplification but worth documenting
- Verify the Solana version bump from v2 to v3 doesn't affect other packages in the monorepo
- The `.so` dumps in `test/programs/dumps/` are binary files committed to git — consider documenting how to regenerate them

.github/actions/setup-anchor/action.yml
```diff
@@ -0,0 +1,63 @@
...
+    default: packages/dynamic-instructions/test/anchor/Cargo.lock
+  build-target-path:
+    description: Path to the build target directory to cache
+    required: false
```
trevor-cortex, line 17: Same path issue here:

```suggestion
    default: packages/dynamic-instructions/test/programs/anchor/target
```

packages/dynamic-client/src/instruction-encoding/arguments/encode-instruction-arguments.ts
```diff
@@ -0,0 +1,115 @@
+import { getNodeCodec, type ReadonlyUint8Array } from '@codama/dynamic-codecs';
+import {
+    CODAMA_ERROR__DYNAMIC_INSTRUCTIONS__ARGUMENT_MISSING,
+    CODAMA_ERROR__DYNAMIC_INSTRUCTIONS__DEFAULT_VALUE_MISSING,
+    CODAMA_ERROR__DYNAMIC_INSTRUCTIONS__FAILED_TO_ENCODE_ARGUMENT,
+    CODAMA_ERROR__UNEXPECTED_NODE_KIND,
+    CodamaError,
+} from '@codama/errors';
```
trevor-cortex, line 8: `Codec` is only used as a type annotation. Use `import type` to avoid pulling it into the runtime bundle:

```suggestion
import { type Codec, mergeBytes } from '@solana/codecs';
```

packages/dynamic-client/src/instruction-encoding/accounts/create-account-meta.ts
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            const finalAddress = isAccountProvided ? toAddress(accountAddressInput) : resolvedAccountAddress;
+
+            // Optional accounts with "programId" strategy: e.g. PMP's setData instruction `buffer` account. (isWritable, isOptional and "programId" strategy).
+            // But when buffer is null it resolves to the program address which cannot be writable, hence must be downgraded to readonly.
+            const isProgramAddress = finalAddress !== null && finalAddress === programAddress;
+            const role = isProgramAddress
+                ? getReadonlyAccountRole(ixAccountNode, signers)
+                : getAccountRole(ixAccountNode, signers);
```
trevor-cortex: Nit: The `isProgramAddress` role downgrade applies to *all* accounts that happen to match the program address, not just those resolved via `optionalAccountStrategy`. If someone explicitly passes the program address for a writable account, it'll get unexpectedly downgraded to readonly. 

This might be intentional (the on-chain runtime forbids writable program accounts anyway), but it could mask bugs where a user provides the wrong address. Consider narrowing this check to only optional accounts that were resolved via strategy, or at least documenting this behavior.

packages/dynamic-client/src/cli/commands/generate-client-types/generate-client-types.ts
```diff
@@ -0,0 +1,333 @@
...
+        }
+        case 'arrayTypeNode':
+        case 'setTypeNode': {
+            const itemType = codamaTypeToTS(type.item, definedTypes);
+            const needsParens = itemType.includes(' | ') || itemType.includes(' & ');
+            return needsParens ? `(${itemType})[]` : `${itemType}[]`;
+        }
+        case 'mapTypeNode': {
+            const v = codamaTypeToTS(type.value, definedTypes);
+            return `Record<string, ${v}>`;
```
trevor-cortex, line 263: Map keys are always typed as `string` regardless of the actual `node.key` type. This works for most IDLs (where map keys are strings), but if a program uses `publicKeyTypeNode` or `numberTypeNode` keys, the generated type won't reflect that. Worth a comment noting this is a known simplification.

packages/dynamic-client/src/instruction-encoding/visitors/pda-seed-value.ts
```diff
@@ -0,0 +1,177 @@
...
+        },
+
+        visitNoneValue: async (_node: NoneValueNode) => await Promise.resolve(new Uint8Array(0)),
+
+        visitNumberValue: async (node: NumberValueNode) => {
+            // Sanity check: a violation here indicates a malformed IDL, not a user input error.
+            if (!Number.isInteger(node.number) || node.number < 0 || node.number > 0xff) {
+                throw new CodamaError(CODAMA_ERROR__DYNAMIC_INSTRUCTIONS__INVARIANT_VIOLATION, {
+                    message: `NumberValueNode PDA seed is out of range: must be a valid u8 (0–255), got ${node.number}`,
+                });
```
trevor-cortex, line 150: This assumes `NumberValueNode` in PDA seeds is always a single u8 byte. That's currently true for all known IDLs (PDA seeds are raw bytes), but the assertion message says "must be a valid u8" without citing why. Consider adding a brief comment explaining that PDA constant seeds using `numberValueNode` are always single-byte values per the Codama spec, so future readers understand this isn't an arbitrary limitation.

packages/dynamic-client/src/cli/commands/generate-client-types/generate-client-types.ts
mikhd: Fixed.

.github/actions/setup-anchor/action.yml
rogaldh: Fixed

packages/dynamic-client/src/instruction-encoding/arguments/encode-instruction-arguments.ts
rogaldh: Done

packages/dynamic-client/src/instruction-encoding/accounts/create-account-meta.ts
rogaldh: Improved

.github/actions/setup-anchor/action.yml
rogaldh: Fixed

lorisleiva (approved):
As discussed offline, reviewing this is extremely challenging so let's get this merge first to unblock the explorer integration and we can make incremental improvements later on.

PR #990, opened 2026-04-09
Resolve nested seed paths in v01 PDA extraction

ioxde:
This is a direct follow-up to #984 which left the nested path handling as a TODO.

PDA seeds with nested paths like `mint.authority` or `args.owner` were skipped. This PR resolves them by looking up the field type from IDL type definitions.

`{ kind: 'account', path: 'mint.authority' }` + a `Mint` type def with `authority: pubkey` -> `variablePdaSeedNode('mintAuthority', publicKeyTypeNode())`

`{ kind: 'arg', path: 'args.owner' }` where `args` is a struct with `owner: pubkey` -> `variablePdaSeedNode('owner', publicKeyTypeNode())`

Program seeds with nested paths still skip (with a warning). `pdaSeedNodeFromAnchorV01` can now return `undefined`.

### DIFF
--- a/packages/dynamic-client/test/programs/anchor/tests/blog.test.ts
+++ b/packages/dynamic-client/test/programs/anchor/tests/blog.test.ts
@@ -108,8 +108,8 @@ describe('blog', () => {
 
             const [profilePda] = await programClient.pdas.profile({ authority: payer });
 
-            // Create post (manual PDA — Codama can't express profile.post_count dependency)
-            const [postPda] = await programClient.pdas.post({ postId: 0, profile: profilePda });
+            // Create post (manual PDA — uses profile.post_count nested seed)
+            const [postPda] = await programClient.pdas.post({ profile: profilePda, profilePostCount: 0 });
             const createPostIx = await programClient.methods
                 .createPost({ content: 'Original content', title: 'Original' })
                 .accounts({ authority: payer, post: postPda })
@@ -139,7 +139,7 @@ describe('blog', () => {
             await ctx.sendInstruction(createProfileIx, [payer]);
 
             const [profilePda] = await programClient.pdas.profile({ authority: payer });
-            const [postPda] = await programClient.pdas.post({ postId: 0, profile: profilePda });
+            const [postPda] = await programClient.pdas.post({ profile: profilePda, profilePostCount: 0 });
 
             const createPostIx = await programClient.methods
                 .createPost({ content: 'World', title: 'Hello' })
@@ -175,7 +175,7 @@ describe('blog', () => {
             await ctx.sendInstruction(createProfileIx, [payer]);
 
             const [profilePda] = await programClient.pdas.profile({ authority: payer });
-            const [postPda] = await programClient.pdas.post({ postId: 0, profile: profilePda });
+            const [postPda] = await programClient.pdas.post({ profile: profilePda, profilePostCount: 0 });
 
             const createPostIx = await programClient.methods
                 .createPost({ content: 'There', title: 'Hi' })
--- a/packages/dynamic-client/test/programs/anchor/tests/blog.test.ts
+++ b/packages/dynamic-client/test/programs/anchor/tests/blog.test.ts
@@ -10,7 +10,7 @@ import { loadIdl, SvmTestContext } from '../../test-utils';
 
 const idl = loadIdl('blog-idl.json');
 const programClient = createProgramClient<BlogProgramClient>(idl);
-const programSoPath = path.resolve(__dirname, '..', '..', 'dumps', 'blog.so');
+const programSoPath = path.resolve(__dirname, '..', 'target', 'deploy', 'blog.so');
 
 describe('blog', () => {
     let ctx: SvmTestContext;
--- a/packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
+++ b/packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
@@ -8,13 +8,15 @@ import {
     InstructionArgumentNode,
     isNode,
     pdaNode,
+    PdaSeedNode,
+    PdaSeedValueNode,
     PdaValueNode,
     pdaValueNode,
     PublicKeyValueNode,
     publicKeyValueNode,
 } from '@codama/nodes';
 
-import { IdlV01InstructionAccount, IdlV01InstructionAccountItem, IdlV01TypeDef } from './idl';
+import { IdlV01InstructionAccount, IdlV01InstructionAccountItem, IdlV01Pda, IdlV01Seed, IdlV01TypeDef } from './idl';
 import { pdaSeedNodeFromAnchorV01 } from './PdaSeedNode';
 import type { GenericsV01 } from './unwrapGenerics';
 
@@ -84,60 +86,10 @@ export function instructionAccountNodeFromAnchorV01(
     const isWritable = idl.writable ?? false;
     const name = prefix ? `${prefix}_${idl.name ?? ''}` : (idl.name ?? '');
     let defaultValue: PdaValueNode | PublicKeyValueNode | undefined;
-
     if (idl.address) {
         defaultValue = publicKeyValueNode(idl.address, name);
     } else if (idl.pda) {
-        const seedResults = idl.pda.seeds.map(seed =>
-            pdaSeedNodeFromAnchorV01(seed, instructionArguments, prefix, idlTypes, generics),
-        );
-
-        if (seedResults.every((r): r is NonNullable<typeof r> => r != null)) {
-            const seedDefinitions = seedResults.map(r => r.definition);
-            const seedValues = seedResults.flatMap(r => (r.value ? [r.value] : []));
-
-            let programId: string | undefined;
-            let programIdValue: AccountValueNode | ArgumentValueNode | undefined;
-            if (idl.pda.program !== undefined) {
-                const result = pdaSeedNodeFromAnchorV01(
-                    idl.pda.program,
-                    instructionArguments,
-                    prefix,
-                    idlTypes,
-                    generics,
-                );
-                if (!result) {
-                    logWarn(`Skipping PDA for account "${name}": program seed could not be resolved.`);
-                    return instructionAccountNode({ defaultValue, docs, isOptional, isSigner, isWritable, name });
-                }
-                if (
-                    isNode(result.definition, 'constantPdaSeedNode') &&
-                    isNode(result.definition.value, 'bytesValueNode') &&
-                    result.definition.value.encoding === 'base58'
-                ) {
-                    programId = result.definition.value.data;
-                } else if (result.value && isNode(result.value.value, ['accountValueNode', 'argumentValueNode'])) {
-                    programIdValue = result.value.value;
-                }
-            }
-
-            const camelName = camelCase(name);
-            const isSelfReferential =
-                seedValues.some(sv => isNode(sv.value, 'accountValueNode') && sv.value.name === camelName) ||
-                (programIdValue != null &&
-                    isNode(programIdValue, 'accountValueNode') &&
-                    programIdValue.name === camelName);
-            if (isSelfReferential) {
-                logWarn(`Skipping PDA for account "${name}": a seed references the account itself.`);
-            }
-            if (!isSelfReferential) {
-                defaultValue = pdaValueNode(
-                    pdaNode({ name, programId, seeds: seedDefinitions }),
-                    seedValues,
-                    programIdValue,
-                );
-            }
-        }
+        defaultValue = resolvePdaDefaultValue(idl.pda, name, instructionArguments, prefix, idlTypes, generics);
     }
 
     return instructionAccountNode({
@@ -148,4 +100,82 @@ export function instructionAccountNodeFromAnchorV01(
         isWritable,
         name,
     });
+}
+
+function resolvePdaDefaultValue(
+    pda: IdlV01Pda,
+    name: string,
+    instructionArguments: InstructionArgumentNode[],
+    prefix: string | undefined,
+    idlTypes: IdlV01TypeDef[],
+    generics: GenericsV01,
+): PdaValueNode | undefined {
+    const seeds = resolveSeeds(pda.seeds, instructionArguments, prefix, idlTypes, generics);
+    if (!seeds) return undefined;
+
+    let programId: string | undefined;
+    let programValue: AccountValueNode | ArgumentValueNode | undefined;
+    if (pda.program) {
+        const result = resolveProgramSeed(pda.program, name, instructionArguments, prefix, idlTypes, generics);
+        if (!result) return undefined;
+        programId = result.id;
+        programValue = result.value;
+    }
+
+    const camelName = camelCase(name);
+    const isSelfReferential =
+        seeds.values.some(sv => isNode(sv.value, 'accountValueNode') && sv.value.name === camelName) ||
+        (programValue != null && isNode(programValue, 'accountValueNode') && programValue.name === camelName);
+    if (isSelfReferential) {
+        logWarn(`Skipping PDA for account "${name}": a seed references the account itself.`);
+        return undefined;
+    }
+
+    return pdaValueNode(pdaNode({ name, programId, seeds: seeds.definitions }), seeds.values, programValue);
+}
+
+function resolveSeeds(
+    seeds: IdlV01Seed[],
+    instructionArguments: InstructionArgumentNode[],
+    prefix: string | undefined,
+    idlTypes: IdlV01TypeDef[],
+    generics: GenericsV01,
+): { definitions: PdaSeedNode[]; values: PdaSeedValueNode[] } | undefined {
+    const results = seeds.map(seed => pdaSeedNodeFromAnchorV01(seed, instructionArguments, prefix, idlTypes, generics));
+    if (!results.every((r): r is NonNullable<typeof r> => r != null)) {
+        return undefined;
+    }
+    return {
+        definitions: results.map(r => r.definition),
+        values: results.flatMap(r => (r.value ? [r.value] : [])),
+    };
+}
+
+function resolveProgramSeed(
+    program: IdlV01Seed,
+    name: string,
+    instructionArguments: InstructionArgumentNode[],
+    prefix: string | undefined,
+    idlTypes: IdlV01TypeDef[],
+    generics: GenericsV01,
+): { id?: string; value?: AccountValueNode | ArgumentValueNode } | undefined {
+    const result = pdaSeedNodeFromAnchorV01(program, instructionArguments, prefix, idlTypes, generics);
+    if (!result) {
+        logWarn(`Skipping PDA for account "${name}": program seed could not be resolved.`);
+        return undefined;
+    }
+
+    if (
+        isNode(result.definition, 'constantPdaSeedNode') &&
+        isNode(result.definition.value, 'bytesValueNode') &&
+        result.definition.value.encoding === 'base58'
+    ) {
+        return { id: result.definition.value.data };
+    }
+
+    if (result.value && isNode(result.value.value, ['accountValueNode', 'argumentValueNode'])) {
+        return { value: result.value.value };
+    }
+
+    return {};
 }
--- a/packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
+++ b/packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
@@ -1,3 +1,4 @@
+import { logWarn } from '@codama/errors';
 import {
     AccountValueNode,
     ArgumentValueNode,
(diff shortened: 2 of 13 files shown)


lorisleiva:
@trevor-cortex

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
+            if (
+                isNode(argumentType, 'sizePrefixTypeNode') &&
+                isNode(argumentType.type, 'stringTypeNode') &&
+                argumentType.type.encoding === 'utf8' &&
+                isNode(argumentType.prefix, 'numberTypeNode') &&
+                argumentType.prefix.format === 'u32'
+            ) {
+                argumentType = stringTypeNode('utf8');
+            }
+
```
trevor-cortex, line 102: Nit: the two return branches at line 102–106 and 109–112 produce the same shape. You could collapse them into a single return after the string-unwrapping logic:

```ts
return {
    definition: variablePdaSeedNode(argumentName, argumentType),
    value: pdaSeedValueNode(argumentName, argumentValueNode(argumentName)),
};
```

The `if (pathParts.length > 1)` guard before this return isn't doing anything differently.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
```diff
@@ -74,38 +88,61 @@ export function instructionAccountNodeFromAnchorV01(
...
-                seedValues,
-                programIdValue,
-            );
+                const camelName = camelCase(name);
+                const isSelfReferential =
+                    seedValues.some(sv => isNode(sv.value, 'accountValueNode') && sv.value.name === camelName) ||
+                    (programIdValue != null &&
+                        isNode(programIdValue, 'accountValueNode') &&
+                        programIdValue.name === camelName);
+                if (isSelfReferential) {
```
trevor-cortex: Ultra-minor: the `if (isSelfReferential)` warning and `if (!isSelfReferential)` assignment could be a single `if/else`:

```ts
if (isSelfReferential) {
    logWarn(...);
} else {
    defaultValue = pdaValueNode(...);
}
```

trevor-cortex (approved):
## Summary

This PR resolves nested seed paths (e.g. `mint.authority`, `args.owner`) in v01 PDA extraction, which were previously left as a TODO and silently skipped. It introduces a `resolveNestedFieldType` helper that walks struct fields (inline or via IDL type definitions) to determine the leaf type. The function `pdaSeedNodeFromAnchorV01` can now return `undefined` when resolution fails, and callers handle this gracefully by skipping the PDA default value.

Additionally, the PR adds self-referential seed detection — if a seed references the account being derived, the PDA default is skipped with a warning.

## Key things to watch

1. **Breaking return type change**: `pdaSeedNodeFromAnchorV01` now returns `| undefined`. This is a well-handled breaking change internally — all call sites check for `undefined` — but any external consumers of this export would need updating.

2. **Self-referential detection**: The new self-referential check (lines 131–138 in `InstructionAccountNode.ts`) correctly catches cases like `vault` referencing itself, and nested paths like `guard.mint` where the account name matches the seed's account reference. The test for `my_guard` with `path: 'my_guard.mint'` is a good edge case.

3. **Nested program paths**: Program seeds with nested paths are still explicitly skipped with a warning — this is a reasonable incremental choice since program seeds with nested paths are rare.

- The `resolveNestedFieldType` helper only handles `structTypeNode` and `definedTypeLinkNode`. Other wrappers (e.g. `optionTypeNode`) would fall through and return `undefined`. This seems fine for now — real-world IDL seeds shouldn't reference optional fields — but worth keeping in mind.
- The `InstructionNode.test.ts` change for the `distribution` account is interesting: it confirms the self-referential detection works at the integration level (the account `distribution` has a seed `path: 'distribution.group_mint'` referencing itself, so the PDA default is correctly omitted).
- Test coverage is thorough: nested account paths, nested arg paths, unresolvable types, self-referential accounts, prefixed nested groups, and program seed edge cases are all covered.

packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
```diff
@@ -291,17 +294,75 @@ test('it correctly prefixes PDA seed account references in nested groups', () =>
...
     const nodes = instructionAccountNodesFromAnchorV01(
-        // [
-        //     accountNode({
-        //         data: sizePrefixTypeNode(
-        //             structTypeNode([structFieldTypeNode({ name: 'authority', type: publicKeyTypeNode() })]),
-        //             numberTypeNode('u32'),
-        //         ),
-        //         name: 'mint',
-        //     }),
-        // ],
```
lorisleiva, line 304: The idea was that these would be uncommented when nested arguments/accounts are supported. Is the idea that the new tests are taking care of that deleted code now?

packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
```diff
@@ -406,17 +596,99 @@ test('it handles PDAs with a program id that points to another account', () => {
...
```
lorisleiva, line 419: Same here

packages/nodes-from-anchor/src/v01/ProgramNode.ts
```diff
@@ -27,7 +27,7 @@ export function programNodeFromAnchorV01(idl: IdlV01): ProgramNode {
         definedTypes,
         errors: errors.map(errorNodeFromAnchorV01),
         events: events.map(event => eventNodeFromAnchorV01(event, types, generics)),
-        instructions: instructions.map(instruction => instructionNodeFromAnchorV01(instruction, generics)),
+        instructions: instructions.map(instruction => instructionNodeFromAnchorV01(instruction, generics, types)),
```
lorisleiva: nit: `instructionNodeFromAnchorV01(instruction, types, generics)`. All other functions order them like that.

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
```diff
@@ -74,38 +88,61 @@ export function instructionAccountNodeFromAnchorV01(
     if (idl.address) {
         defaultValue = publicKeyValueNode(idl.address, name);
     } else if (idl.pda) {
```
lorisleiva, line 91: This is starting to have way too much code for an if-statement. Would you mind extracting this to a local private function so we can understand this better?

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
             const prefixedAccountName = prefix ? `${prefix}_${accountName}` : accountName;
+
+            if (pathParts.length > 1) {
+                const accountTypeName = seed.account ?? accountName;
+                const rootType = typeNodeFromAnchorV01({ defined: { name: accountTypeName } }, generics);
+                const resolved = resolveNestedFieldType(rootType, pathParts.slice(1), idlTypes, generics);
+                if (!resolved) {
+                    logWarn(`Could not resolve nested account path "${seed.path}" for PDA seed.`);
+                    return undefined;
+                }
```
lorisleiva, line 54: I wonder if here we should just throw? Before we were silently ignoring nested paths because that was a valid scenario we were not able to support. Now that we are, if the nested seed don't resolve properly than it is a malformation on the IDL itself right?

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
         }
         case 'arg': {
-            // Ignore nested paths.
-            const [originalArgumentName] = seed.path.split('.');
-            const argumentName = camelCase(originalArgumentName);
-            const argumentNode = instructionArguments.find(({ name }) => name === argumentName);
-            if (!argumentNode) {
-                throw new CodamaError(CODAMA_ERROR__ANCHOR__ARGUMENT_TYPE_MISSING, { name: originalArgumentName });
+            const pathParts = seed.path.split('.');
+            const argumentName = camelCase(pathParts.length > 1 ? pathParts[pathParts.length - 1] : pathParts[0]);
```
lorisleiva: nit: this is equivalent to:


```suggestion
            const argumentName = camelCase(pathParts[pathParts.length - 1]);
```

Because `pathParts.length - 1 === 0` when `pathParts.length === 1`.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
 
-            // Anchor uses unprefixed strings for PDA seeds even though the
-            // argument itself uses a Borsh size-prefixed string. Thus, we
-            // must recognize this case and convert the type accordingly.
-            const isBorshString =
-                isNode(argumentNode.type, 'sizePrefixTypeNode') &&
-                isNode(argumentNode.type.type, 'stringTypeNode') &&
-                argumentNode.type.type.encoding === 'utf8' &&
-                isNode(argumentNode.type.prefix, 'numberTypeNode') &&
-                argumentNode.type.prefix.format === 'u32';
```
lorisleiva, line 62: I'd rather you kept the `isBorshString` variable so we know what this massive if condition checks.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
             };
         }
         default:
             throw new CodamaError(CODAMA_ERROR__ANCHOR__SEED_KIND_UNIMPLEMENTED, { kind });
     }
 }
+
+function resolveNestedFieldType(
+    rootType: TypeNode,
+    fieldPath: string[],
```
lorisleiva: nit: Keep your naming consistent.

```suggestion
    pathParts: string[],
```

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
+    idlTypes: IdlV01TypeDef[],
+    generics: GenericsV01,
+): TypeNode | undefined {
+    let currentType = rootType;
+
+    for (const fieldName of fieldPath) {
+        const target = camelCase(fieldName);
+
+        if (isNode(currentType, 'structTypeNode')) {
```
lorisleiva, line 132: Do we also need to worry about `tupleTypeNodes`? Is having a path like `foo.0.bar` a thing?

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
+                value: pdaSeedValueNode(argumentName, argumentValueNode(argumentName)),
             };
         }
         default:
             throw new CodamaError(CODAMA_ERROR__ANCHOR__SEED_KIND_UNIMPLEMENTED, { kind });
     }
 }
+
+function resolveNestedFieldType(
+    rootType: TypeNode,
```
lorisleiva: I wouldn't call that `rootType` since there is a `RootNode` which makes this confusing.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,115 @@ export function pdaSeedNodeFromAnchorV01(
...
+            const linkName = currentType.name;
+            const typeDef = idlTypes.find(t => camelCase(t.name) === linkName);
+            if (!typeDef || typeDef.type.kind !== 'struct' || !typeDef.type.fields) return undefined;
+            const { fields } = typeDef.type;
+            if (!fields.length || typeof fields[0] !== 'object' || !('name' in fields[0])) return undefined;
+            const field = (fields as IdlV01Field[]).find(f => camelCase(f.name) === target);
+            if (!field) return undefined;
+            currentType = typeNodeFromAnchorV01(field.type, generics);
+            continue;
+        }
```
lorisleiva: Could you not prefix this before the `structTypeNode` condition and do something like this instead;

```ts
if (isNode(currentType, 'definedTypeLinkNode')) {
    const linkName = currentType.name;
    const typeDef = idlTypes.find(t => camelCase(t.name) === linkName);
    if (!typeDef) return undefined;
    currentType = typeNodeFromAnchorV01(typeDef, generics);
}

// Now the `structTypeNode` condition (and any others in the future) can pick up from that fetched linked node.
```

packages/nodes-from-anchor/test/v01/InstructionNode.test.ts
```diff
@@ -34,23 +34,13 @@ test('it creates instruction nodes', () => {
...
-                    // defaultValue: pdaValueNode(
-                    //     pdaNode({
-                    //         name: 'distribution',
-                    //         seeds: [
-                    //             constantPdaSeedNodeFromBytes('base58', 'F9bS'),
-                    //             variablePdaSeedNode('distributionGroupMint', publicKeyTypeNode()),
-                    //         ],
-                    //     }),
-                    //     [pdaSeedValueNode("distributionGroupMint", accountValueNode('distribution', 'group_mint'))],
-                    // ),
```
lorisleiva, line 53: Same here

packages/nodes-from-anchor/test/v01/pdaSeedNode.test.ts
```diff
@@ -17,24 +20,86 @@ import { pdaSeedNodeFromAnchorV01 } from '../../src';
 test('it creates a PdaSeedNode from a const Anchor seed', () => {
     const nodes = pdaSeedNodeFromAnchorV01({ kind: 'const', value: [11, 57, 246, 240] }, []);
 
-    expect(nodes.definition).toEqual(constantPdaSeedNodeFromBytes('base58', 'HeLLo'));
-    expect(nodes.value).toBeUndefined();
+    expect(nodes!.definition).toEqual(constantPdaSeedNodeFromBytes('base58', 'HeLLo'));
+    expect(nodes!.value).toBeUndefined();
```
lorisleiva: If we throw instead of returning undefined, we remove the need for these casts.

lorisleiva (changes requested):

packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
ioxde: The commented code assumed account data would be passed directly via `accountNode`, but nested types are now resolved through `idlTypes` instead.

packages/nodes-from-anchor/test/v01/InstructionAccountNode.test.ts
ioxde: Same as above

packages/nodes-from-anchor/src/v01/InstructionAccountNode.ts
ioxde: Good idea. This comment let me also catch a mistake [if a program seed has a nested path](https://github.com/ioxde/codama/commit/10ab1a84e7e6e39e9fd5b3280b7f44e43076ddb1) being skipped.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
ioxde: This was changed in a37264c. It was changed to `pathParts.slice(1).join('_')` to avoid collisions so the ternary is needed now. The rebase onto 394e9b6a exposed the issue.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
ioxde: `parentType` okay?

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
ioxde: Ended up changing this with feedback from above

packages/nodes-from-anchor/test/v01/InstructionNode.test.ts
ioxde: Same feedback about using `idlTypes` as above.

packages/nodes-from-anchor/test/v01/pdaSeedNode.test.ts
ioxde: I switched all tests to `?.`.  Unsure about the `throw`. See above comment about throw vs log.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
ioxde: Honestly not sure. I know Anchor PDAs aren't perfect. I think the best path would be to throw by default (with a detailed message pointing to docs/what to fix) but allowing a user to override per error. I am not sure that whole change fits in this PR though. Seems more like a project-level standard / convention?

I personally do like the fail-first override behavior because it forces the user to understand their decisions. Console warnings don't always tell the real severity of the issue / easy to accidentally ignore.

I'll wait for more direction to make changes here.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
ioxde: Great catch b0c4d41b1b39a85eca83c93349afa1b9035b3568

ioxde:
I think I addressed all comments outside of the `logWarn` vs `throw` question. Let me know if you have any other questions / concerns.

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
lorisleiva: I'm happy with keeping this as-is if you're concerned about malformed Anchor IDLs.

packages/nodes-from-anchor/test/v01/pdaSeedNode.test.ts
```diff
@@ -17,24 +21,151 @@ import { pdaSeedNodeFromAnchorV01 } from '../../src';
...
+test('it resolves nested path through tuple then struct (foo.0.bar)', () => {
+    const nodes = pdaSeedNodeFromAnchorV01({ kind: 'arg', path: 'foo.0.bar' }, [
+        instructionArgumentNode({
+            name: 'foo',
+            type: tupleTypeNode([structTypeNode([structFieldTypeNode({ name: 'bar', type: numberTypeNode('u8') })])]),
+        }),
+    ]);
+
+    expect(nodes?.definition).toEqual(variablePdaSeedNode('0Bar', numberTypeNode('u8')));
+    expect(nodes?.value).toEqual(pdaSeedValueNode('0Bar', argumentValueNode('0Bar')));
```
lorisleiva, line 148: This worries me a bit because most languages will refuse to render a variable that starts with a numerical value. I wonder if we should just use the whole path as the seed name. So here it would become `foo0Bar`. That way, it would technically not be possible for a variable to start with a number because we always start with the array of `InstructionArgumentNodes`.

lorisleiva (changes requested):
Thank you for all the changes and sorry for the late re-review.

Just one last little thing that caught my attention and after that this is good to go. Thank you for your patience. 🙏

packages/dynamic-client/test/programs/anchor/tests/blog.test.ts
```diff
@@ -10,7 +10,7 @@ import { loadIdl, SvmTestContext } from '../../test-utils';
 
 const idl = loadIdl('blog-idl.json');
 const programClient = createProgramClient<BlogProgramClient>(idl);
-const programSoPath = path.resolve(__dirname, '..', '..', 'dumps', 'blog.so');
+const programSoPath = path.resolve(__dirname, '..', 'target', 'deploy', 'blog.so');
```
mikhd, line 13: Hi @ioxde,

I think it should be resolved `.so` from dumps.

As recently we've [get rid](https://github.com/codama-idl/codama/pull/1005) of custom anchor programs building in tests and instead tests use pre-build committed binaries from dumps.

Thank you!

packages/nodes-from-anchor/src/v01/PdaSeedNode.ts
```diff
@@ -34,40 +40,111 @@ export function pdaSeedNodeFromAnchorV01(
...
         }
         case 'arg': {
-            // Ignore nested paths.
-            const [originalArgumentName] = seed.path.split('.');
-            const argumentName = camelCase(originalArgumentName);
-            const argumentNode = instructionArguments.find(({ name }) => name === argumentName);
-            if (!argumentNode) {
-                throw new CodamaError(CODAMA_ERROR__ANCHOR__ARGUMENT_TYPE_MISSING, { name: originalArgumentName });
+            const pathParts = seed.path.split('.');
+            const argumentName = camelCase(pathParts.length > 1 ? pathParts.slice(1).join('_') : pathParts[0]);
```
mikhd, line 69: Could we double-check this, please?

The arg branch drops the root, so two different struct arguments sharing a field name within the same instruction always collide.

Lets say two different InstructionArgument inputs have the same keys.
If I understood this correctly, PdaSeed depending either on `input1.owner` or `input2.owner` will get the same `owner` name. So it will become impossible to identify pda seed dependency.

```
[
    instructionArgumentNode({
        name: 'input1',
        type: structTypeNode([
            structFieldTypeNode({ name: 'owner', type: publicKeyTypeNode() }),
            structFieldTypeNode({ name: 'amount', type: numberTypeNode('u64') }),
        ]),
    }),
    instructionArgumentNode({
        name: 'input2',
        type: structTypeNode([
            structFieldTypeNode({ name: 'owner', type: publicKeyTypeNode() }),
            structFieldTypeNode({ name: 'amount', type: numberTypeNode('u64') }),
        ]),
    }),
]
```

PR #991, opened 2026-04-11
Guard against undefined events in getAllEvents and identifyData

ioxde:
I ran into this issue after rebasing my fork to 35516d4b0757456dbcd52473d275f9361348b308. If you want me to merge the commits into one let me know.

`getAllEvents` returns `[undefined]` and `identifyData` throws when `program.events` is undefined. ProgramNodes from JSON deserialization or pre-events codama trees won't have the field.

The visitors already guard with `(node.events ?? [])` but I also added tests for them.

NOTE: Same gap exists in all `getAll*` functions. They all do bare `program.X` without `?? []`.

I think the root cause is `rootNode()`. It passes programs through without calling `programNode()`, so raw JSON objects keep `undefined` for missing fields like `events`. The `?? []` guards are the safe incremental fix. A deeper fix would be normalizing through `programNode()` in `rootNode()`, but unsure if that has side-effects

Happy to update this PR if you want

### DIFF
--- a/.changeset/guard-undefined-events.md
+++ b/.changeset/guard-undefined-events.md
@@ -1,0 +1,6 @@
+---
+"@codama/nodes": patch
+"@codama/dynamic-parsers": patch
+---
+
+Guard against undefined `events` on ProgramNode in `getAllEvents` and `identifyData`.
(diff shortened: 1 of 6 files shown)


ioxde:
I ran into this issue again even with these changes. I think the correct fix is either normalizing the data structure as mentioned above or changing the events type to `readonly events?: TEvents;` so it is correctly typed and callers will know.

Closing this- let me know if you'd like me to follow up with a different PR.

PR #998, opened 2026-05-18
feat(nodes-from-anchor): add event support to V00 anchor IDLs

stegaBOB:
Old (V00) anchor IDLs define events inline with fields and rely on a SHA256("event:EventName") discriminator. Bring V00 in line with V01 by emitting EventNodes for these IDLs.

### DIFF
--- a/.changeset/v00-anchor-events.md
+++ b/.changeset/v00-anchor-events.md
@@ -1,0 +1,5 @@
+---
+'@codama/nodes-from-anchor': minor
+---
+
+Add event support to V00 anchor IDLs
--- a/packages/nodes-from-anchor/src/discriminators.ts
+++ b/packages/nodes-from-anchor/src/discriminators.ts
@@ -19,3 +19,9 @@ export const getAnchorAccountDiscriminatorV00 = (idlName: string): BytesValueNod
     const hash = sha256(bytes as Uint8Array).slice(0, 8);
     return bytesValueNode('base16', hex(hash));
 };
+
+export const getAnchorEventDiscriminatorV00 = (idlName: string): BytesValueNode => {
+    const bytes = getUtf8Codec().encode(`event:${pascalCase(idlName)}`);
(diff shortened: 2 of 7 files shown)


stegaBOB:
Could probably make an argument for this to be a patch bump? Minor seems like a lot. I don't have a preference one way or the other

lorisleiva:
@trevor-cortex

trevor-cortex (approved):
Brings V00 anchor IDLs in line with V01 by emitting `EventNode`s, using the classic `sha256("event:<PascalCase>")[0..8]` discriminator scheme. The shape of the produced node (hidden-prefix struct + constant discriminator) matches `eventNodeFromAnchorV01` exactly, which is what we want.

**Things that look right:**
- Origin gating in `ProgramNode.ts` (`origin === 'anchor' ? … : []`) mirrors how `accountNodeFromAnchorV00` only attaches anchor discriminators when `origin === 'anchor'`. Correct — shank IDLs reusing the V00 schema shouldn't get anchor event semantics.
- 8-byte fixed-size constant matches Anchor's classic prefix.
- `IdlV00EventField` is structurally compatible with what `structTypeNodeFromAnchorV00` reads (`name` + `type`); the `index` flag is effectively ignored, which matches actual on-chain behavior for V00 anchor events (no real indexing in the emitted bytes).
- Changeset is a `minor` bump on a pre-1.0 package — appropriate for a user-facing additive feature.

**Notes for subsequent reviewers:**
- Worth a quick sanity check that the expected discriminator `60b8c5f38b025a94` for `MyEvent` matches what anchor itself emits — easy to verify against any real V00 anchor IDL with events (e.g. an older Metaplex program).
- No "event type missing" error path like V01 has, but that's by design: V00 defines event fields inline, so there's no separate type def to look up.

LGTM.

lorisleiva (approved):
Thank you!

PR #1005, opened 2026-06-04
chore(dynamic client): anchor custom program tests optimization

mikhd:
## Summary

Removes the Rust/Solana/Anchor toolchain from CI which used to be required for `@codama/dynamic-client` tests.
The custom  Anchor test programs are no longer built in CI — instead their prebuilt `.so` binaries are committed to be used in tests. Anchor version updated to `v1.0.2`.

## What changed

- Removed Solana/Rust/Anchor from CI/CD.
- Removed program building from `@codama/dynamic-client` tests.
- Committed custom Anchor programs binaries.
- Added scripts for Anchor builds sync (copy dumps, produce hashes).
  - Recompute each program's source hash and asserts it matches the committed `artifacts/anchor-build-sync-hashes.json`. If the program source changes without a rebuild, the test fails with guidance to run `pnpm anchor:sync:build` (A cheap drift test with the committed artifacts against falling out of sync sources should mitigates possible discrepancy between program code and program in tests)
- Updated test programs `Anchor@1.0.2`.


lorisleiva (approved):
Thanks!

PR #1008, opened 2026-06-13
fix(cli): allow convert command to work without package.json

senzenn:
- Add @codama/nodes-from-anchor as a dependency of @codama/cli
- Try direct import before falling back to the install flow

This lets users run codama convert without needing a package.json.

### DIFF
--- a/.changeset/fix-convert-package-json.md
+++ b/.changeset/fix-convert-package-json.md
@@ -1,0 +1,5 @@
+---
+"@codama/cli": patch
+---
+
+Allow convert command to work without package.json
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -41,6 +41,7 @@
     },
     "dependencies": {
         "@codama/nodes": "workspace:*",
+        "@codama/nodes-from-anchor": "workspace:*",
         "@codama/visitors": "workspace:*",
         "@codama/visitors-core": "workspace:*",
         "commander": "^14.0.2",
--- a/packages/cli/src/utils/nodes.ts
+++ b/packages/cli/src/utils/nodes.ts
@@ -12,6 +12,17 @@ export async function getRootNodeFromIdl(idl: unknown): Promise<RootNode> {
         return idl;
     }
 
+    // Try direct import first — works without a package.json.
+    try {
+        const rootNodeFromAnchor = await importModuleItem<(idl: unknown) => RootNode>({
+            from: '@codama/nodes-from-anchor',
+            item: 'rootNodeFromAnchor',
+        });
+        return rootNodeFromAnchor(idl);
+    } catch {
+        // Module not available via direct import.
+    }
+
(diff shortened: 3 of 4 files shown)


senzenn:
## Before 

Screenshot 2026-06-13 at 8 29 02 PM

## After
Screenshot 2026-06-13 at 9 08 26 PM

lorisleiva:
Hi there, sorry but this behaviour is by design. It decouples the CLI from the exact version of the Anchor adapter you need and allows you to update one without affecting the other. This is the same pattern we follow with renderers so you can use the latest CLI whilst having full control over the version of the renderers that make sense for your project. When running the command you are modifying, if the CLI consumer does not have the Anchor adapter package, the CLI will prompt the user and ask them if they wish for the CLI to install the latest version on their behalf which is the expected behaviour.

senzenn:
Hey @lorisleiva thanks for the reminder on the intentions, and agreed on the decoupling. I've dropped the hard dependency I'd added, so the CLI stays decoupled and the
  install prompt is untouched.

  One thing the prompt doesn't cover though: when `convert` runs without a `package.json`, the prompt never fires. `installMissingDependencies` → `getPackageJson` throws first:

  ```ts
  // packages/cli/src/utils/packageJson.ts:19-21
  if (!(await canRead(packageJsonPath))) {
      throw new CliError('Cannot read package.json.', ...);
  }

  The only remaining change tries to resolve an already-installed adapter before that path:

  // packages/cli/src/utils/nodes.ts:15-24
  try {
      const rootNodeFromAnchor = await importModuleItem(...);
      return rootNodeFromAnchor(idl);
  } catch { /* fall through to install prompt */ }

  So: adapter resolvable → used directly (works with no package.json); not resolvable → existing prompt, unchanged. No new dependency. Acceptable, or would you prefer this
  handled differently?

lorisleiva:
IMO, there are two ways we can make this more `npx` friendly:
- We can add a small CLI directly inside `@codama/nodes-from-anchor`.
- We can try and make `installMissingDependencies` more clever so that it recognises when it's being called in a global environment and prompt to add the dependency to the user's global dependencies as well (which should only need to happen the first time).

senzenn:
Thanks @lorisleiva. I’d lean towards option 1 a small CLI inside ‎⁠@codama/nodes-from-anchor⁠. It keeps the adapter fully decoupled, needs no ‎⁠package.json⁠ or global state, and ‎⁠codama convert⁠ could delegate to it when the adapter isn’t already resolvable. Option 2’s global/npx detection is fragile across package managers, and the “first time only” install breaks when the npx cache is evicted.

Separately: without a ‎⁠package.json⁠, ‎⁠installMissingDependencies⁠ → ‎⁠getPackageJson⁠ throws before the prompt can fire, so the install UX never runs. Happy to PR that fix and prototype the adapter-side CLI if option 1 sounds good.

PR #1009, opened 2026-06-16
fix(cli): allow convert to work without a package.json

senzenn:
Follow-up to #1008 (closed), trimmed down to just the decoupled fix we talked about there.

Running `codama convert` outside a Node project failed with `Cannot read package.json`. The install prompt never got a chance to run because `installMissingDependencies` reads the package.json first and throws when it's missing.

Changes:
- `tryGetPackageJson` no longer throws when there's no package.json. A missing file just means "no deps" and "no package manager set".
- If `@codama/nodes-from-anchor` is already installed, use it directly instead of going through the install prompt. So convert works without a package.json when the adapter is there.

Left out on purpose:
- No hard dependency on `@codama/nodes-from-anchor` (dropped from #1008). It stays decoupled.
- No auto-creating a package.json, no npx detection. The prompt behaves the same as before.

The adapter-side CLI (option 1 from the thread) isn't here, can do that once we agree on the direction.

`tsc --noEmit`, `eslint`, `prettier --check` and `vitest` all pass.


packages/cli/src/utils/packageJson.ts
```diff
@@ -13,19 +13,36 @@ type PackageJson = {
...
 let packageJson: PackageJson | undefined;
-export async function getPackageJson(): Promise<PackageJson> {
-    if (!packageJson) {
+let packageJsonChecked = false;
+
+/**
+ * Reads the local `package.json` if one exists. Returns `undefined` when none is found so callers
+ * can degrade gracefully (e.g. when running outside a Node project or via `npx`).
+ */
+export async function tryGetPackageJson(): Promise<PackageJson | undefined> {
```
lorisleiva, line 22: I'm happy with this direction. 👌

packages/cli/src/utils/nodes.ts
```diff
@@ -12,6 +12,21 @@ export async function getRootNodeFromIdl(idl: unknown): Promise<RootNode> {
...
+        rootNodeFromAnchor = await importModuleItem<(idl: unknown) => RootNode>({
+            from: '@codama/nodes-from-anchor',
+            item: 'rootNodeFromAnchor',
+        });
+    } catch {
+        // Adapter not resolvable directly; fall back to the install flow below.
+    }
+    if (rootNodeFromAnchor) {
+        return rootNodeFromAnchor(idl);
+    }
```
lorisleiva: I might be missing something but I'm confused as to how this solve the problem for `npx codama convert`. If someone calls this for the first time, they won't have `@codama/nodes-from-anchor` installed even globally right? Could you please expand on this for me?

packages/cli/src/utils/nodes.ts
```diff
@@ -12,6 +12,21 @@ export async function getRootNodeFromIdl(idl: unknown): Promise<RootNode> {
...
+    // Resolve an already-installed adapter first — this works without a package.json.
+    // Only the import is guarded: a conversion error from the adapter must surface, not trigger an install.
+    let rootNodeFromAnchor: ((idl: unknown) => RootNode) | undefined;
+    try {
+        rootNodeFromAnchor = await importModuleItem<(idl: unknown) => RootNode>({
+            from: '@codama/nodes-from-anchor',
+            item: 'rootNodeFromAnchor',
+        });
+    } catch {
+        // Adapter not resolvable directly; fall back to the install flow below.
```
trevor-cortex: The bare `catch {}` swallows **any** error from `importModuleItem`, not just "module not found". If the adapter is installed but broken (e.g. a syntax error in a transitive load, or the `rootNodeFromAnchor` export goes missing in a future version), the user gets the generic install prompt instead of the real error, which would be a pretty confusing debugging experience.

The comment above says "Only the import is guarded: a conversion error from the adapter must surface, not trigger an install" — but that's only true *after* the import succeeds. The import itself has both "not resolvable" and "resolvable but failed to load" failure modes, and both currently fall through to the install path.

Probably fine to leave as-is for this PR (it matches the prior behavior of "if anything goes wrong, offer to install"), but if you wanted to tighten it, narrowing the catch to a module-resolution error (e.g. inspect `cause.code === 'ERR_MODULE_NOT_FOUND'` / `'MODULE_NOT_FOUND'`) would make the failure modes more honest. Worth at least a quick mention in the comment that load errors also fall through.

packages/cli/src/utils/nodes.ts
```diff
@@ -20,7 +35,7 @@ export async function getRootNodeFromIdl(idl: unknown): Promise<RootNode> {
         throw new CliError('Cannot proceed without Anchor IDL support.');
     }
 
-    const rootNodeFromAnchor = await importModuleItem<(idl: unknown) => RootNode>({
+    rootNodeFromAnchor = await importModuleItem<(idl: unknown) => RootNode>({
         from: '@codama/nodes-from-anchor',
         item: 'rootNodeFromAnchor',
```
trevor-cortex: Minor: the `let` + reassignment dance reads a bit awkwardly now that the same `importModuleItem` call happens in two places. A tiny helper would flatten it:

```ts
async function tryImportRootNodeFromAnchor() {
    try {
        return await importModuleItem<(idl: unknown) => RootNode>({
            from: '@codama/nodes-from-anchor',
            item: 'rootNodeFromAnchor',
        });
    } catch {
        return undefined;
    }
}
```

Then the body becomes `const fn = await tryImportRootNodeFromAnchor() ?? (await installAndImport()); return fn(idl);`. Not worth blocking on — purely a readability nit.

packages/cli/src/utils/packageJson.ts
```diff
@@ -13,19 +13,36 @@ type PackageJson = {
...
+    if (!json) {
+        const packageJsonPath = resolveRelativePath('package.json');
+        throw new CliError('Cannot read package.json.', [`${pico.bold('Path')}: ${packageJsonPath}`]);
+    }
+    return json;
+}
+
 export async function getPackageJsonDependencies(options: { includeDev?: boolean } = {}): Promise<string[]> {
-    const packageJson = await getPackageJson();
+    const packageJson = await tryGetPackageJson();
```
trevor-cortex, line 45: Worth calling out that this is a quiet semantic change beyond "don't throw on missing package.json": previously `getPackageJsonDependencies` would propagate the `Cannot read package.json` error; now it silently returns `[]`.

For the current caller (`installMissingDependencies`), this is exactly what you want — "no package.json" should mean "all required deps are missing" so the install prompt kicks in. But if any future caller relies on `getPackageJsonDependencies` to assert that a project exists, they'd be surprised. The JSDoc on `tryGetPackageJson` covers the philosophy; might be worth a one-liner on `getPackageJsonDependencies` too saying it returns `[]` outside a Node project.

packages/cli/src/utils/packageJson.ts
```diff
@@ -13,19 +13,36 @@ type PackageJson = {
...
+export async function tryGetPackageJson(): Promise<PackageJson | undefined> {
+    if (!packageJsonChecked) {
         const packageJsonPath = resolveRelativePath('package.json');
-        if (!(await canRead(packageJsonPath))) {
-            throw new CliError('Cannot read package.json.', [`${pico.bold('Path')}: ${packageJsonPath}`]);
+        if (await canRead(packageJsonPath)) {
+            packageJson = await readJson<PackageJson>(packageJsonPath);
         }
-        packageJson = await readJson<PackageJson>(packageJsonPath);
+        packageJsonChecked = true;
```
trevor-cortex, line 28: Small caching subtlety: if `readJson` throws (e.g. corrupt JSON), `packageJsonChecked` stays `false` and the next call retries. That matches the prior behavior of the original `getPackageJson` (which also gated on `packageJson` being set), so no regression — just noting it in case you ever want to make "corrupt package.json" a sticky failure rather than a re-attempt on every call. No action needed.

trevor-cortex (commented):
Trimmed-down follow-up to #1008 — much nicer scope. The fix is two clean parts:

1. `tryGetPackageJson()` returns `undefined` instead of throwing, and `getPackageJsonDependencies` / `detectPackageManagerFromPackageJson` now degrade to "no deps" / "unknown PM" rather than blowing up. `getPackageJson()` is preserved as the throwing variant for callers that genuinely require it.
2. `getRootNodeFromIdl` tries to import `@codama/nodes-from-anchor` directly before falling back to the install prompt, so a globally/npx-resolvable adapter works without a `package.json` in cwd.

The behavior change is small and well-scoped to fixing the `codama convert` UX outside a Node project. Two things worth a look before merging — both around the `nodes.ts` try/catch — plus one note on a subtle behavior shift in `getPackageJsonDependencies`. Nothing blocking from my side.

**Things to verify:**
- Manual smoke test: `codama convert` in a tmp dir with no `package.json`, with `@codama/nodes-from-anchor` reachable both (a) via a globally-linked install and (b) via `npx codama convert`. Both should resolve the adapter without hitting the install prompt.
- Same command in a tmp dir with no `package.json` and **no** adapter reachable — should still fall through to the install prompt and not crash with the old `Cannot read package.json` error.
- `getPackageManager()` in a no-`package.json` dir should now fall through to lockfile / installed-CLI detection / `npm` fallback without throwing.

**Notes for subsequent reviewers:**
- No tests added; the utils don't seem to have a great harness for fs-touching code today, so probably fine — but a unit test for `tryGetPackageJson` returning `undefined` would be cheap insurance if you have a fixture pattern in mind.
- Option 1 from the #1008 thread (adapter-side CLI) is deliberately not in scope here, as called out in the description.

packages/cli/src/utils/nodes.ts
senzenn: You're right, the direct import doesn't help a first-run npx convert that's not what it's for.

  The real fix is tryGetPackageJson not throwing. Before, installMissingDependencies → getPackageJsonDependencies → getPackageJson() read package.json and
  threw Cannot read package.json before the prompt could run. Now a missing file just means "no deps", so it reaches the install prompt instead of
  crashing.

  The direct import is only a shortcut for when the adapter is already resolvable (global install, or hoisted next to the CLI) skip the prompt. When it's
  not resolvable (the npx case), the catch swallows it and we fall through to the same prompt as before.

  So:
  - no package.json, adapter there → converts, no prompt
  - no package.json, adapter missing → install prompt (used to crash here)
  - package.json present → unchanged

  Tested it: packed the CLI and installed it outside the monorepo so nodes-from-anchor is genuinely unresolvable. Cold start now hits Install dependencies?
  (Y/n) (or prints the manual command under CI=1) instead of the old crash.

  Can drop the direct-import block if you'd rather keep this scoped to just the tryGetPackageJson fix.

senzenn:
hey @trevor-cortex Scope's exactly the intent. Ran the smoke tests:

  - no package.json + adapter resolvable → converts directly, no prompt. Tested via the adapter resolvable next to the CLI, same branch a global link / npx
  adapter hits.
  - no package.json + no adapter → packed the CLI and installed it outside the monorepo so nodes-from-anchor is genuinely unresolvable. Falls through to
  Install dependencies? (Y/n), no Cannot read package.json crash. Under CI=1 prints the manual command and exits clean.
  - getPackageManager() with no package.json → confirmed by the above, didn't throw, resolved a PM via the fallback path.

  Unit test for tryGetPackageJson → undefined is fair; can add one in this PR before merge if you want, else follow-up. Adapter-side CLI stays out of scope.

packages/cli/src/utils/nodes.ts
lorisleiva: Thanks for clarifying! Could you please expand on what the `no package.json, adapter missing` scenario does when you type `Y`? Does it create a new package.json locally with the `@codama/nodes-from-anchor` package installed?

packages/cli/src/utils/nodes.ts
senzenn: Good point. Narrowed the catch to module-resolution errors only (`ERR_MODULE_NOT_FOUND` / `MODULE_NOT_FOUND`, walking the cause chain); everything else rethrows. An installed-but-broken adapter now surfaces its real error instead of the install prompt.

packages/cli/src/utils/nodes.ts
senzenn: Done — extracted `tryImportRootNodeFromAnchor()` and `installRootNodeFromAnchor()`. Body is now just `const fn = (await tryImportRootNodeFromAnchor()) ?? (await installRootNodeFromAnchor()); return fn(idl);`.

packages/cli/src/utils/packageJson.ts
senzenn: Added a one-liner: Lists declared dependencies, returning [] outside a Node project (no package.json).

packages/cli/src/utils/packageJson.ts
senzenn: Agreed, that's intentional and matches the old getPackageJson gating

packages/cli/src/utils/nodes.ts
senzenn: Yeah. On Y, the CLI runs the detected package manager (npm/pnpm install @codama/nodes-from-anchor, or yarn add) in the cwd. With no package.json, that's  standard "add to a bare dir" behavior: it creates a package.json there with the adapter as a dependency (plus node_modules, and a lockfile for pnpm/yarn), then the re-import resolves it and convert finishes.

lorisleiva: Gotcha! I do think it is a bit weird that an `npx codama convert` command may end up creating a `package.json`.

I did a bit of research and it looks like we can force an `npx` command to include more than one package by using the following syntax:

```
npx -p codama -p @codama/nodes-from-anchor codama convert
```

We still need the changes from your PRs to allow this to work. I just wonder if we should tweak the logic a little bit such that:

- If `package.json` exists, follow existing logic.
- Else, try to load `@codama/nodes-from-anchor` anyway (as it could be available from npx's temporary install path when using `-p <package>`).
- Else, throw an error explaining that no `package.json` was found and that the following dependencies are not available.

We can even construct the `npx -p codama -p <package1> -p <package2> codama ...` command in the error message to help the user run the command without ending up with an unwanted `package.json`.

What do you think?

senzenn: Agreed, creating a package.json silently is bad like real bad. The PR already covers the -p case though it resolves an already installed @codama/nodes-from-anchor   before trying to install, so npx -p codama -p @codama/nodes-from-anchor  codama convert finds it on npx's tmp path. 
So the only changes is the fallback when  there is no package.json  and the adpater  isn't resolvable, throw message instead of just blindly installing . 
if package.json exists: Then install prompt (current  one )
2. No package.json adapter found -> use it 
3. No package.json , not found  -> error listing the missing deps + npx -p codama -p <pkg>. codama command  that rerun, just wondering if you look into this.

lorisleiva: Yeah that's what I had in mind! ☺️

senzenn: hey @lorisleiva  I've implemented this flow.  I also tests for each scenario and verified the npx -p flow without creating a package.json.

packages/cli/src/utils/nodes.ts
```diff
@@ -1,30 +1,75 @@
...
+        throw error;
+    }
+}
+
+function isMissingModuleError(error: unknown, moduleName: string): boolean {
+    for (let current: unknown = error; current != null; current = (current as { cause?: unknown }).cause) {
+        const { code, message } = current as { code?: unknown; message?: unknown };
+        const isModuleNotFound = code === 'ERR_MODULE_NOT_FOUND' || code === 'MODULE_NOT_FOUND';
+        const namesRequestedModule =
+            typeof message === 'string' && (message.includes(`'${moduleName}'`) || message.includes(`"${moduleName}"`));
```
lorisleiva: nit: I wonder if we need this extra assertion since we're only ever trying to import that module anyway.

lorisleiva (approved):
Looks good, thanks!

EDIT: One weird test failing in CI haha. If you wouldn't mind taking care of that? 🙏

lorisleiva:
@trevor-cortex

trevor-cortex:
Found it — it's `nodes.test.ts › suggests npx without installing when package.json and the adapter are missing`. The test asserts plain strings:

```ts
'Missing dependency: @codama/nodes-from-anchor',
...
'Re-run with: npx -p codama -p @codama/nodes-from-anchor codama convert anchor.json codama.json',
```

but the production code wraps those labels with `pico.bold(...)` / `pico.yellow(...)`. Picocolors' `isColorSupported` is true whenever `process.env.CI` is set (even without a TTY), so in GitHub Actions those calls actually emit ANSI escapes — `\x1b[1mMissing dependency\x1b[22m: …` — and the `toEqual([...])` comparison fails. Locally there's no `CI` env var and no TTY, so the same calls return plain strings and the test passes. Classic env-dependent flake.

Cleanest fix is to mock picocolors in that test file so the assertions are deterministic regardless of where they run:

```ts
vi.mock('picocolors', () => ({
    default: { bold: (s: string) => s, yellow: (s: string) => s },
}));
```

(`pico` is imported as the default export in `nodes.ts`, so mocking `default` is what matters.) Alternatively, strip ANSI from the actual values before comparing, but the mock is less code and keeps the expected strings readable.

No other tests look affected — `packageJson.test.ts` doesn't touch picocolors output, and the other `nodes.test.ts` cases only assert on `error.message` (no `pico` calls on that path).

packages/cli/src/utils/nodes.ts
senzenn: pushed again @lorisleiva

packages/cli/src/utils/nodes.ts
senzenn: Yeah my bad. Dropped the message check; any ERR_MODULE_NOT_FOUND/MODULE_NOT_FOUND now just means the adapter isn't resolvable and we fall back. Updated  the test to match. Pushed in de081ad. @lorisleiva

packages/cli/src/utils/nodes.ts
```diff
@@ -1,30 +1,72 @@
...
     if (isRootNode(idl)) {
         return idl;
     }
 
+    const rootNodeFromAnchor =
+        (await tryImportRootNodeFromAnchor()) ?? (await resolveRootNodeFromAnchor(options.npxCommandArgs));
+    return rootNodeFromAnchor(idl);
+}
+
+async function resolveRootNodeFromAnchor(npxCommandArgs: string[] = ['convert']): Promise<RootNodeFromAnchor> {
```
lorisleiva: nit: to avoid weird behaviour in other parts of the CLI, I'd either make `npxCommandArgs` required or give it a more generic default like `['[...]']` so that it displays:

```
Re-run with: npx -p codama -p @codama/nodes-from-anchor codama [...]
```

When the arguments are not provided.

lorisleiva (approved):
Thanks! Just one last nit and LGTM. 👌

packages/cli/src/utils/nodes.ts
senzenn: ['convert'] default produced a misleading codama convert suggestion when running other commands. Changed the default to ['[...]'] as you suggested, so it now shows Re-run with: npx -p codama -p @codama/nodes-from-anchor codama [...].

PR #1010, opened 2026-06-17
feat(dynamic-parsers): fallback to identify data for single node kind without discriminator

mikhd:
## Summary

`parseInstruction` / `parseInstructionData` returns `undefined` for single-instruction programs that declare no discriminator, e.g. [SPL Memo v4](https://explorer.solana.com/address/Memo4c2pN8afCj432Lb7RMVKi9PbQnnW7ewFFaV3oAH/idl) `Memo4c2pN8afCj432Lb7RMVKi9PbQnnW7ewFFaV3oAH`. Its instruction data is just a UTF-8 string without instruction selector.

Current ix identification is discriminator-based, so parsing fails when discriminator doesn't exist.

## What changed

- Updated `identifyData` to fall back fallback to identify data for single node kind without discriminator.
- Added tests.

### DIFF
--- a/packages/dynamic-parsers/src/parsers.ts
+++ b/packages/dynamic-parsers/src/parsers.ts
@@ -37,7 +37,19 @@ export function parseInstructionData(
     root: RootNode,
     bytes: ReadonlyUint8Array | Uint8Array,
 ): ParsedData<InstructionNode> | undefined {
-    return parseData(root, bytes, 'instructionNode');
+    const parsed = parseData(root, bytes, 'instructionNode');
+    if (parsed) return parsed;
+
+    // Single-instruction programs may omit a discriminator (e.g. Memo program).
+    // parseData returns undefined due to being unable to identify it by discriminator
+    // Decode it directly.
+    const [singleIx, ...rest] = root.program.instructions;
+    if (singleIx && rest.length === 0 && !singleIx.discriminators?.length) {
+        const path = [root, root.program, singleIx] as NodePath<InstructionNode>;
+        return { data: getNodeCodec(path).decode(bytes), path };
+    }
+
+    return undefined;
 }
 
 export function parseData<TKind extends ParsableNodeKind>(
--- a/packages/dynamic-parsers/src/parsers.ts
+++ b/packages/dynamic-parsers/src/parsers.ts
@@ -40,7 +40,7 @@ export function parseInstructionData(
     const parsed = parseData(root, bytes, 'instructionNode');
     if (parsed) return parsed;
 
-    // Single-instruction programs may omit a discriminator (e.g. Memo program).
+    // Single-instruction programs may omit a discriminator (e.g. `Memo4c2pN8afCj432Lb7RMVKi9PbQnnW7ewFFaV3oAH`).
     // parseData returns undefined due to being unable to identify it by discriminator
     // Decode it directly.
     const [singleIx, ...rest] = root.program.instructions;
--- a/packages/dynamic-parsers/src/parsers.ts
+++ b/packages/dynamic-parsers/src/parsers.ts
@@ -37,19 +37,7 @@ export function parseInstructionData(
     root: RootNode,
     bytes: ReadonlyUint8Array | Uint8Array,
 ): ParsedData<InstructionNode> | undefined {
-    const parsed = parseData(root, bytes, 'instructionNode');
-    if (parsed) return parsed;
-
-    // Single-instruction programs may omit a discriminator (e.g. `Memo4c2pN8afCj432Lb7RMVKi9PbQnnW7ewFFaV3oAH`).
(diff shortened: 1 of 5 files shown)


packages/dynamic-parsers/src/parsers.ts
```diff
@@ -37,7 +37,19 @@ export function parseInstructionData(
     root: RootNode,
     bytes: ReadonlyUint8Array | Uint8Array,
 ): ParsedData<InstructionNode> | undefined {
-    return parseData(root, bytes, 'instructionNode');
+    const parsed = parseData(root, bytes, 'instructionNode');
+    if (parsed) return parsed;
+
+    // Single-instruction programs may omit a discriminator (e.g. Memo program).
```
rogaldh: Maybe it is worth mentioning the exact version of a program. 4th if I am not mistaken

rogaldh (approved):
LGTM

packages/dynamic-parsers/src/parsers.ts
mikhd: Sure! Added address to make it discoverable, just in case https://github.com/codama-idl/codama/pull/1010/commits/c590433e981e02992186f68ca3b556b34016c95f

lorisleiva (commented):
Thanks! I like this approach but reading the PR I had an idea worth discussing.

What if instead we updated the `identifyData` part of the parser such that:
- It first goes through nodes that have discriminators.
- If none was identified it tries to grab the first node that does not have any discriminators (i.e. the fallback).

Pros:
- We can have a mixture of discriminated and non-discriminated nodes to define fallbacks.
- Now we get that logic for free for events and accounts.
- The logic goes near the "identification" part of the process which feels right.

Cons:
- Multiple fallback nodes means the first one gets picked => but that feels intuitive to me.
- Using `parseData` directly instead of `parseInstructionData`/`parseAccountData`/`parseEventData` means you _could_ identify an account instead of an instruction if they both have non-discriminated nodes => but that feels like the expected behaviour to me if you don't use the more specific helpers as you're just trying to understand "something" from the program. An alternative would be for these to identify multiple nodes but that would be a significant change to the API.

Lmk what you think. 🙏

mikhd:
> Thanks! I like this approach but reading the PR I had an idea worth discussing.
> 
> What if instead we updated the `identifyData` part of the parser such that:
> 
> * It first goes through nodes that have discriminators.
> * If none was identified it tries to grab the first node that does not have any discriminators (i.e. the fallback).
> 
> Pros:
> 
> * We can have a mixture of discriminated and non-discriminated nodes to define fallbacks.
> * Now we get that logic for free for events and accounts.
> * The logic goes near the "identification" part of the process which feels right.
> 
> Cons:
> 
> * Multiple fallback nodes means the first one gets picked => but that feels intuitive to me.
> * Using `parseData` directly instead of `parseInstructionData`/`parseAccountData`/`parseEventData` means you _could_ identify an account instead of an instruction if they both have non-discriminated nodes => but that feels like the expected behaviour to me if you don't use the more specific helpers as you're just trying to understand "something" from the program. An alternative would be for these to identify multiple nodes but that would be a significant change to the API.
> 
> Lmk what you think. 🙏

Hey @lorisleiva,
Thanks for feedback!

I like your suggestion and, actually, my first approach was more or less the same - initially I wanted to tweak `identify.ts` layer.

But I was looking into `getByteIdentificationVisitor` and realized that if required behaviour is `"When Instruction is single AND discriminator doesn't exist THEN identify such Instruction"` then I would want this logic to exist within `visitInstruction` branch only and it was not possible since I needed to know how many Instructions does `Root` have.
Also seeing that current behaviour of identification process is discriminator-matching, I decided to keep it untouched and so ended up patching parse layer for our specific single-ix case.

**Questions if we patch identifyData:**
1) My thoughts were that we keep this fallback only when given Node without discriminator is single. If we have multiple Nodes of the same kind without discriminator, then I gravitate more to returning `undefined` instead of using fallback for picking the first one. (I'm fine with any decision here, but just wanted to align)

2) `could identify an account instead of an instruction if they both have non-discriminated nodes`.
We could also consider other option (not sure if it's better or not, though):
To mitigate such ambiguity, we could enforce some stricter rules, i.e. `"When 'kind' argument is undefined OR 'kind'.length > 1 THEN do not fallback"` (e.g. `parseData(root, [1,2,3])` or `parseData(root, [1,2,3], ['accountNode', 'eventNode'])`).

Basically both notes are about - when we can't clearly identify Node, return `undefined`.

Before I jump into implementation, please let me know what do you think about these? 🙏

mikhd:
Updated `identifyData` with aforementioned fallback!

lorisleiva (approved):
Perfect! Thank you!

rogaldh (approved):
LGTM

PR #1012, opened 2026-06-18
feat(dynamic-address-resolution): refactor pda-seed-value visitor

mikhd:
### Summary

This is [4] PR for dynamic-client refactoring. It refactors internals of pda-seed-value visitor (the public API stays unchanged).
Extract the constant-seed part of `createPdaSeedValueVisitor` into a standalone visitor to simplify `resolveStandalonePda` method. The pda-seed-value visitor now extends the constant-pda-seed with variable-seed handling (`visitAccountValue` / `visitArgumentValue`) on top.

###  Change

- Added `createConstantPdaSeedValueVisitor` for the 9 constant value-node kinds (boolean / bytes / constant / none / number / programId / publicKey / some / string).
- Added `resolveConstantPdaSeedValue` resolver (internal) — facade over the visitor.
- Refactored `createPdaSeedValueVisitor` to extend the constant base via `extendVisitor` core helper, keeping only the `visitAccountValue`, `visitArgumentValue` overrides for variable seeds.
- Updated `resolveStandalonePda` to use `resolveConstantPdaSeedValue` directly. Drops the STANDALONE_IX_NODE stub and the manual visitOrElse dispatch.
- Added tests.

### DIFF
--- a/packages/dynamic-address-resolution/src/visitors/pda-seed-value.ts
+++ b/packages/dynamic-address-resolution/src/visitors/pda-seed-value.ts
@@ -58,10 +58,12 @@ type PdaSeedValueSupportedNodeKind = (typeof PDA_SEED_VALUE_SUPPORTED_NODE_KINDS
 
 export type ConstantPdaSeedValueVisitorContext = {
     programId: Address;
-    seedTypeNode?: TypeNode;
 };
 
-export type PdaSeedValueVisitorContext = BaseResolutionContext & ConstantPdaSeedValueVisitorContext;
+export type PdaSeedValueVisitorContext = BaseResolutionContext &
+    ConstantPdaSeedValueVisitorContext & {
+        seedTypeNode?: TypeNode;
+    };
 
 /**
  * Visitor for resolving PdaSeedValueNode value to raw bytes.
@@ -77,7 +79,7 @@ export function createPdaSeedValueVisitor(
     const accountsInput = ctx.accountsInput ?? {};
     const argumentsInput = ctx.argumentsInput ?? {};
 
-    const base = createConstantPdaSeedValueVisitor({ programId, seedTypeNode });
+    const base = createConstantPdaSeedValueVisitor({ programId });
 
     const visitor: Visitor<Promise<ReadonlyUint8Array>, PdaSeedValueSupportedNodeKind> = extendVisitor(base, {
         visitAccountValue: async (node: AccountValueNode) => {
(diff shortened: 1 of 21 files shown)


packages/dynamic-address-resolution/src/visitors/pda-seed-value.ts
```diff
@@ -121,6 +132,44 @@ export function createPdaSeedValueVisitor(
...
+
+/**
+ * Base PDA seed value visitor that handles constant seed kinds only
+ * (boolean / bytes / constant / none / number / programId / publicKey / some / string).
+ * The account/argument handlers throw by default and are meant to be extended if needed (e.g. for variable seeds).
+ */
+export function createConstantPdaSeedValueVisitor(
+    ctx: ConstantPdaSeedValueVisitorContext,
+): Visitor<Promise<ReadonlyUint8Array>, PdaSeedValueSupportedNodeKind> {
+    const { programId } = ctx;
```
rogaldh, line 162: It looks like `seedTypeNode` has no use here

packages/dynamic-address-resolution/src/visitors/pda-seed-value.ts
```diff
@@ -121,6 +132,44 @@ export function createPdaSeedValueVisitor(
...
+
+    const visitor: Visitor<Promise<ReadonlyUint8Array>, PdaSeedValueSupportedNodeKind> = {
+        // Throws error by default since constant seeds should not depend on accounts.
+        visitAccountValue: async (node: AccountValueNode) => {
```
rogaldh, line 166: What do you think if we add a test where `visitAccountValue` and `visitArgumentValue` throw? As we might have recursive calls within it would be better to rely on the test.

rogaldh (approved):
LGTM

lorisleiva (approved):
Thanks!

packages/dynamic-address-resolution/src/visitors/pda-seed-value.ts
mikhd: Great catch, fixed! https://github.com/codama-idl/codama/pull/1012/commits/41f1ba8cb0e2d22943d4b2bf9463da63eec2e1fc

packages/dynamic-address-resolution/src/visitors/pda-seed-value.ts
mikhd: Yeah, we already have few tests for those cases:
- https://github.com/codama-idl/codama/pull/1012/changes#diff-fee52c21d96369c5c18d51d6a1398a14e554679e9347b310950a5fd8627c8f82R7
- https://github.com/codama-idl/codama/pull/1012/changes#diff-0832fc130303413922f1d1125d72875210a7d0d44b41a4d7d9e4b33f7edcdcc9R8
- https://github.com/codama-idl/codama/pull/1012/changes#diff-e6a83f7f581bbf150e1e36d01cc17f4e84b83c0a85d306d3fce82179fdc61e12R40
- https://github.com/codama-idl/codama/pull/1012/changes#diff-e4e1c2ec89093a615efd8a673d7b0b89535a5a8de9a6e8a1e09df4a87ef924dbR40

PR #1013, opened 2026-06-22
Integrate @codama/spec@1.7.0

lorisleiva:
Integrates `@codama/spec@1.7.0-rc.0` into the TS codegen pipeline and regenerates `@codama/node-types`, `@codama/nodes`, and `@codama/visitors-core` with the new presentation layer (display nodes, provide/inject, account-field selector).

Held against `main` for CI and review but not merged until the spec stabilises at `1.7.0` — letting us bump to later RCs here as the spec evolves. No changeset is included for the same reason.

Key changes:
- `anyNode` TypeExpr: both `typeExpr.ts` fragments render it as the top-level `Node` type; the visitor generator visits-and-asserts it against `REGISTERED_NODE_KINDS`.
- Config maps (`CATEGORY_DIRECTORIES`, `GENERIC_PARAM_ORDER`, walk-order maps, `NODE_CONFIGS`) absorb the new `display`/`provides` attributes; a `paramName`-only `AttributeOverride` variant renames `providedNode.node` so it doesn't shadow the visitor's `node` parameter.
- Input-type generator now widens every `stringIdentifier()` attribute to `string` (not just `name`).
- Visitor fixes: `getByteSizeVisitor`/`getMaxByteSizeVisitor` add `visitStringType => null` now that `stringTypeNode` is no longer a leaf; `getDebugStringVisitor` drops the trailing `()` for childless nodes; `dynamic-codecs` adds `visitInjectedValue` that throws (injected values must be resolved before codec time).
- Tests: 11 new per-node fixtures for the new spec kinds, plus `anyNode` unit tests.

Lint, build, and test pass across all 17 packages.

PR #1014, opened 2026-06-25
Add display value formatting and provide/inject resolution

lorisleiva:
This PR begins the clear-signing display-text feature (sRFC 39) with its value layer. It adds resolveInjectedValue, which resolves the provide/inject graph — literal values, injected keys (with fallback), account references, and account fields read via an optional fetchAccountData hook returning Kit's decoded Account — within an explicit DisplayResolutionContext. On top of it, it adds the per-display-node value formatters: scaled token amounts (treating decimals as correctness, so an unresolvable scale yields no output rather than a misleading value, while an unresolvable unit merely drops the suffix), ISO 8601 date-times, HH:mm:ss durations, and sliced strings. These are internal building blocks and are not yet exported from the package; the resolution context is kept explicit so the value layer can be exercised in isolation ahead of the interpolation, fallback-list, and orchestration PRs. No changeset is included, consistent with the rest of the 1.7.0 stack.

### DIFF
--- a/packages/dynamic-instructions/src/display/format-value.ts
+++ b/packages/dynamic-instructions/src/display/format-value.ts
@@ -1,0 +1,97 @@
+import type {
+    AmountNumberDisplayNode,
+    DateTimeNumberDisplayNode,
+    DurationNumberDisplayNode,
+    StringDisplayNode,
+} from 'codama';
+
+import { resolveInjectedValue } from './resolve-injected-value';
+import type { DisplayResolutionContext } from './types';
+
+/**
+ * Formats an integer as a scaled amount with an optional unit (e.g. `1100000000` → `"1.1 SOL"`).
+ *
+ * The scale (`decimals`) is treated as correctness: when it cannot be resolved, this returns
+ * `null` so callers fall back to the raw value rather than display a misscaled — and therefore
+ * misleading — amount. The `unit` is treated as enrichment: when it cannot be resolved, the
+ * scaled value is returned without a suffix.
+ */
+export async function formatAmountValue(
+    value: bigint | number,
+    node: AmountNumberDisplayNode,
+    context: DisplayResolutionContext,
+): Promise<string | null> {
+    const decimals = node.decimals ? await resolveInjectedValue(node.decimals, context) : 0;
+    if (decimals === null || (typeof decimals !== 'bigint' && typeof decimals !== 'number')) {
+        return null;
+    }
+
+    const scaled = scaleByDecimals(value, Number(decimals));
+    if (scaled === null) return null;
+
+    const unit = node.unit ? await resolveInjectedValue(node.unit, context) : null;
+    return typeof unit === 'string' && unit !== '' ? `${scaled} ${unit}` : scaled;
+}
+
+/**
+ * Formats an integer counting ticks since the Unix epoch as an ISO 8601 date-time string.
+ * `ticksPerSecond` (default `1`) converts the raw ticks back to seconds.
+ */
+export function formatDateTimeValue(value: bigint | number, node: DateTimeNumberDisplayNode): string | null {
+    const seconds = toSeconds(value, node.ticksPerSecond);
+    if (seconds === null) return null;
+    const date = new Date(seconds * 1000);
+    if (Number.isNaN(date.getTime())) return null;
+    return date.toISOString();
+}
+
+/**
+ * Formats an integer counting ticks as an elapsed duration in `HH:mm:ss`.
+ * `ticksPerSecond` (default `1`) converts the raw ticks back to seconds.
+ */
+export function formatDurationValue(value: bigint | number, node: DurationNumberDisplayNode): string | null {
+    const totalSeconds = toSeconds(value, node.ticksPerSecond);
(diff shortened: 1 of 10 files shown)


lorisleiva:
* **#1019** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1019?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1016** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1016?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1015** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1015?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1014** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1014?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1014?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1013** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1013?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/dynamic-instructions/src/display/resolve-injected-value.ts
```diff
@@ -0,0 +1,86 @@
...
+ * - `numberValueNode` / `stringValueNode`: the literal value.
+ * - `injectedValueNode`: looks the key up in `provides`, resolving the matched provider's node;
+ *   when no provider supplies the key, falls back to the injection's own `fallback`.
+ * - `accountValueNode`: the referenced account's address.
+ * - `accountFieldValueNode`: a field of the referenced account's decoded data, fetched via
+ *   `fetchAccountData`.
+ *
+ * Returns `null` when the value cannot be resolved so callers can fall back safely.
+ */
+export async function resolveInjectedValue(
```
mikhd, line 30: **question**: is argument-sourced injection intentionally scoped out?

What made me unsure was that [collectReferencedMembers](https://github.com/codama-idl/codama/pull/1016/changes#diff-142a365bed7ec850dbed6edb865a66992a79aabf54281710063694e3c98308c3R70) does handle `argumentValueNode` (added in PR#1016) but `resolveInjectedValue` does not. The fix would look like just `if (isNode(node, 'argumentValueNode')) return toResolvedValue(context.data[node.name]);`, if i'm not mistaken.

Structure example:
```js
instructionNode({
  arguments: [
    instructionArgumentNode({ name: 'amount', type: numberTypeNode('u64', 'le', {
      display: amountNumberDisplayNode({ decimals: injectedValueNode({ key: 'decimals' }) }),
    })}),
    instructionArgumentNode({ name: 'decimals', type: numberTypeNode('u8') }),
  ],
  provides: [ providedNode('decimals', argumentValueNode('decimals')) ], // <-- argument-sourced provide
})
```

And something like these:
```js
test('it resolves an injected value provided by an argumentValueNode', async () => {
    // Given an injected key provided by an argumentValueNode.
    const node = injectedValueNode({ key: 'decimals' });
    const provides = providesMap(providedNode('decimals', argumentValueNode('decimals')));

    // When we resolve it (the decoded argument is present in context.data).
    const result = await resolveInjectedValue(node, context({ data: { decimals: 6 }, provides }));

    // Then we expect the argument value.
    expect(result).toBe(6);
});

test('it resolves an argumentValueNode', async () => {
    // Given an argument value node resolved directly.
    const node = argumentValueNode('decimals');

    // When we resolve it.
    const result = await resolveInjectedValue(node, context({ data: { decimals: 6 } }));

    // Then we expect the argument value.
    expect(result).toBe(6);
});
```

packages/dynamic-instructions/src/display/format-value.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Divides an integer value by `10 ^ decimals`, trimming trailing zeros from the fractional part.
+ * Returns `null` for negative decimals, which cannot describe a fixed-point scale.
+ */
+function scaleByDecimals(value: bigint | number, decimals: number): string | null {
+    if (!Number.isInteger(decimals) || decimals < 0) return null;
+    if (decimals === 0) return value.toString();
+
+    const negative = value < 0;
+    const digits = (negative ? -BigInt(value) : BigInt(value)).toString().padStart(decimals + 1, '0');
```
mikhd, line 83: Should we also double-guard that `value: bigint | number` number is not float? Otherwise conversion to BigInt will throw RangeError.

mikhd (approved):

lorisleiva:
### Merge activity

* **Jul 9, 2:00 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/codama/1014).
* **Jul 9, 2:03 PM UTC**: [Graphite](https://app.graphite.com/github/pr/codama-idl/codama/1014) rebased this pull request as part of a merge.
* **Jul 9, 2:03 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/codama-idl/codama/1014).

PR #1015, opened 2026-06-25
Add interpolated-intent and fallback-list render modes

lorisleiva:
This PR adds the two clear-signing render modes on top of the value layer (sRFC 39). interpolateIntent renders an instruction's interpolatedIntent template by substituting flat ${data.<argument>} and ${accounts.<account>} placeholders, returning null when the template is absent or a referenced name cannot be resolved so the caller can fall back. listFallback builds the structured fallback list of labelled fields for the instruction's arguments and accounts, honouring skip strategies (including whenInjected), label overrides, and struct flattening with prefixes. Both consume a new formatArgumentValue bridge that maps a decoded value to its type's display node — dispatching to the amount, date-time, duration, and string formatters, resolving linked enums to their variant labels, and degrading to a raw form when no display metadata applies. Defined-type links are followed through a single shared resolveDisplayType helper, and the whole display layer threads one flat DisplayContext (instruction, decoded data, account addresses, provide/inject graph, account fetching, and link resolution) so every piece stays unit-testable ahead of the orchestration PR. These remain internal building blocks, not yet exported from the package. No changeset is included, consistent with the rest of the 1.7.0 stack.

### DIFF
--- a/packages/dynamic-instructions/src/display/format-argument-value.ts
+++ b/packages/dynamic-instructions/src/display/format-argument-value.ts
@@ -1,0 +1,112 @@
+import {
+    type EnumTypeNode,
+    isNode,
+    isScalarEnum,
+    pascalCase,
+    resolveNestedTypeNode,
+    titleCase,
+    type TypeNode,
+} from 'codama';
+
+import { isObjectRecord } from '../shared/util';
+import { formatAmountValue, formatDateTimeValue, formatDurationValue, formatStringValue } from './format-value';
+import type { DisplayContext } from './types';
+
+/**
+ * Formats a single decoded value according to the presentation metadata on its type.
+ *
+ * Numbers, strings, and enum variants are rendered through their value-display nodes when
+ * present; `definedTypeLinkNode`s are followed via the context's link resolver so linked
+ * enums resolve to their variants. Any value without applicable display metadata — and any
+ * value whose formatter cannot resolve its inputs — falls back to a raw string form.
+ */
+export async function formatArgumentValue(
+    type: TypeNode,
+    value: unknown,
+    displayContext: DisplayContext,
+): Promise<string> {
+    const resolvedType = resolveDisplayType(type, displayContext);
+
+    if (isNode(resolvedType, 'numberTypeNode') && resolvedType.display && isNumeric(value)) {
+        const formatted = await formatNumber(resolvedType.display, value, displayContext);
+        if (formatted !== null) return formatted;
+    }
+
+    if (isNode(resolvedType, 'stringTypeNode') && resolvedType.display && typeof value === 'string') {
+        return formatStringValue(value, resolvedType.display);
+    }
+
+    if (isNode(resolvedType, 'enumTypeNode')) {
+        return formatEnumValue(resolvedType, value);
+    }
+
+    return rawValue(value);
+}
+
+/**
+ * Resolves nested type wrappers and follows a single `definedTypeLinkNode` to its underlying type.
+ * Shared by the value formatter and the fallback list so links resolve identically in both.
+ */
+export function resolveDisplayType(type: TypeNode, displayContext: DisplayContext): TypeNode {
+    const resolved = resolveNestedTypeNode(type);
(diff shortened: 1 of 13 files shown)


lorisleiva:
* **#1019** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1019?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1016** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1016?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1015** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1015?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1015?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1014** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1014?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1013** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1013?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/dynamic-instructions/src/display/format-argument-value.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/** Renders a value without any display metadata as a safe, human-readable string. */
+function rawValue(value: unknown): string {
+    if (typeof value === 'string') return value;
+    if (typeof value === 'bigint' || typeof value === 'number' || typeof value === 'boolean') {
+        return value.toString();
+    }
+    if (isObjectRecord(value) && typeof value.__kind === 'string') return titleCase(value.__kind);
+    return JSON.stringify(value, (_key, v: unknown) => (typeof v === 'bigint' ? v.toString() : v));
```
mikhd, line 107: Just in case, there is [safeStringify](https://github.com/codama-idl/codama/blob/main/packages/dynamic-instructions/src/shared/util.ts#L28) in shared/util.

However, not sure it will be acceptable in this case to return `non-serializable <value_type>` instead of throwing and error when JSON.stringify fails.

packages/dynamic-instructions/src/display/list-fallback.ts
```diff
@@ -0,0 +1,83 @@
...
+    });
+}
+
+/**
+ * Determines whether a member is hidden from the fallback list given its `skip` strategy.
+ * `'always'` always hides; `'whenInjected'` hides when a provider exposes the member's name
+ * (its value is surfaced elsewhere through the provide/inject graph); `'never'`/absent shows.
+ */
+function isSkipped(
+    skip: 'always' | 'never' | 'whenInjected' | undefined,
```
mikhd: If i'm not mistaken, `DisplaySkip` could be imported from `import { type DisplaySkip } from 'codama';`
```suggestion
    skip: DisplaySkip | undefined,
```

packages/dynamic-instructions/src/display/format-argument-value.ts
mikhd, line 107: I think when `value: unknown` is `undefined`, then `JSON.stringify(undefined)` would return `undefined` instead of desired `string`.

I guess, this could happen when some key of context data is set to `undefined` - i.e.
`displayContext { data: { decimals: undefined } }`

mikhd (approved):

lorisleiva:
### Merge activity

* **Jul 9, 2:00 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/codama-idl/codama/1015).
* **Jul 9, 2:05 PM UTC**: [Graphite](https://app.graphite.com/github/pr/codama-idl/codama/1015) rebased this pull request as part of a merge.
* **Jul 9, 2:05 PM UTC**: @lorisleiva merged this pull request with [Graphite](https://app.graphite.com/github/pr/codama-idl/codama/1015).

PR #1018, opened 2026-07-02
feat(cli): add additionalIdls config option

### LINKED ISSUE #983: Add single command to generate clients for multiple programs within an anchor workspace
I can only generate clients one at a time for each program in an Anchor workspace. Each of them would have a separate codama.json config. I would like codama to detect multiple programs in an Anchor workspace and then generate clients for all of them with a single command and a single codama.json file at the root.

If this is already possible, I would like to document that somewhere in the README or docs. If not, I'm happy to take on this issue if the maintainers agree that this is a good idea.

plutohan:
Closes #983.

Implements the `additionalIdls` approach from your comment @lorisleiva: the config can list additional Anchor IDLs, and their programs get merged into the root node's `additionalPrograms`, so one `codama.json` generates clients for every program in an Anchor workspace.

Includes a README entry, a test, and a changeset.

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -71,6 +71,7 @@ Here are the available options for the `run` command:
 The Codama configuration file defines an object containing the following fields:
 
 - `idl` (string): The path to the IDL file. This can be a Codama IDL or an Anchor IDL which will be automatically converted to a Codama IDL.
+- `additionalIdls` (array): An array of paths to additional Anchor IDLs. Their programs are included alongside the main IDL's program (as `additionalPrograms`), so a single config can generate clients for multiple programs in an Anchor workspace.
     ```json
     {
         "idl": "path/to/idl"
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -71,7 +71,7 @@ Here are the available options for the `run` command:
 The Codama configuration file defines an object containing the following fields:
 
 - `idl` (string): The path to the IDL file. This can be a Codama IDL or an Anchor IDL which will be automatically converted to a Codama IDL.
-- `additionalIdls` (array): An array of paths to additional Anchor IDLs. Their programs are included alongside the main IDL's program (as `additionalPrograms`), so a single config can generate clients for multiple programs in an Anchor workspace.
+- `additionalIdls` (array): An array of paths to additional IDLs. Their programs are included alongside the main IDL's program (as `additionalPrograms`), so a single config can generate clients for multiple programs in an Anchor workspace.
     ```json
     {
         "idl": "path/to/idl"
--- a/packages/cli/src/parsedConfig.ts
+++ b/packages/cli/src/parsedConfig.ts
@@ -45,33 +45,25 @@ async function parseConfig(
     options: Pick<ProgramOptions, 'idl'>,
 ): Promise<ParsedConfig> {
     const idlPath = parseIdlPath(config, configPath, options);
-    const idlContent = await importModuleItem({ identifier: 'IDL', from: idlPath });
-    const rootNode = await mergeAdditionalIdls(idlContent, config.additionalIdls ?? [], configPath);
+    const additionalIdlPaths = (config.additionalIdls ?? []).map(idl => resolveConfigPath(idl, configPath));
+
+    // Load and convert the main IDL and every additional IDL the same way, in parallel.
+    const idls = await Promise.all([
+        loadIdl(idlPath, 'IDL'),
+        ...additionalIdlPaths.map(path => loadIdl(path, 'additional IDL')),
+    ]);
+
+    const [mainIdl] = idls;
+    const rootNode = mergeRootNodes(idls.map(idl => idl.rootNode));
     const scripts = parseScripts(config.scripts ?? {}, configPath);
     const visitors = (config.before ?? []).map((v, i) => parseVisitorConfig(v, configPath, i, null));
 
-    return { configPath, idlContent, idlPath, rootNode, scripts, before: visitors };
+    return { configPath, idlContent: mainIdl.content, idlPath, rootNode, scripts, before: visitors };
 }
 
-async function mergeAdditionalIdls(
-    idlContent: unknown,
-    additionalIdls: readonly string[],
-    configPath: string | null,
-): Promise<RootNode> {
-    // Convert the main IDL and every additional IDL to a root node in parallel,
-    // keeping the main root node first.
-    const rootNodes = await Promise.all([
-        getRootNodeFromIdl(idlContent),
-        ...additionalIdls.map(async additionalIdl => {
-            const additionalIdlPath = resolveConfigPath(additionalIdl, configPath);
-            const additionalIdlContent = await importModuleItem({
-                identifier: 'additional IDL',
-                from: additionalIdlPath,
-            });
-            return await getRootNodeFromIdl(additionalIdlContent);
-        }),
-    ]);
-    return mergeRootNodes(rootNodes);
+async function loadIdl(from: string, identifier: string): Promise<{ content: unknown; rootNode: RootNode }> {
+    const content = await importModuleItem({ identifier, from });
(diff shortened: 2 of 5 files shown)


packages/cli/src/parsedConfig.ts
```diff
@@ -46,13 +46,34 @@ async function parseConfig(
...
+
+    const additionalPrograms = [...mainRootNode.additionalPrograms];
+    for (const additionalIdl of additionalIdls) {
+        const additionalIdlPath = resolveConfigPath(additionalIdl, configPath);
+        const additionalIdlContent = await importModuleItem({ identifier: 'additional IDL', from: additionalIdlPath });
+        const additionalRootNode = await getRootNodeFromIdl(additionalIdlContent);
+        additionalPrograms.push(...getAllPrograms(additionalRootNode));
+    }
+
+    return rootNode(mainRootNode.program, additionalPrograms);
```
mikhd: I think the version could be lost here. (worth checking other fields as well) 

`getRootNodeFromIdl` returns a `RootNode` one of two ways:
  - If the IDL already is a Codama root node, it returns it as-is, preserving the whatever specified version.
  - Otherwise it's an Anchor IDL, and `rootNodeFromAnchor(idl)` builds a root node via the `rootNode()`.

packages/cli/test/parsedConfig.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    getConfigMock.mockResolvedValue([{ additionalIdls: ['a.json', 'b.json'], idl: 'main.json' }, '/root/codama.json']);
+    getRootNodeFromIdlMock
+        .mockResolvedValueOnce(rootNodeWith('main'))
+        .mockResolvedValueOnce(rootNodeWith('a'))
+        .mockResolvedValueOnce(rootNodeWith('b'));
+
+    const parsed = await getParsedConfig({});
+
+    expect(parsed.rootNode.program.name).toBe('main');
+    expect(parsed.rootNode.additionalPrograms.map(p => p.name)).toEqual(['a', 'b']);
```
mikhd, line 37: I would double check that `version` and other fields are preserved after `additionalPrograms` merge.
Thank you!

plutohan:
Good catch, the merge rebuilt the root node and reset version and standard. It now spreads the main root node instead, with test assertions for both.

aleksola (approved):

aleksola (approved):

packages/cli/src/parsedConfig.ts
```diff
@@ -46,13 +46,35 @@ async function parseConfig(
 ): Promise<ParsedConfig> {
     const idlPath = parseIdlPath(config, configPath, options);
     const idlContent = await importModuleItem({ identifier: 'IDL', from: idlPath });
-    const rootNode = await getRootNodeFromIdl(idlContent);
+    const mainRootNode = await getRootNodeFromIdl(idlContent);
+    const rootNode = await mergeAdditionalIdls(mainRootNode, config.additionalIdls ?? [], configPath);
```
lorisleiva: I would rather we implemented this as:
1. Promise.all await all conversions from IDL string to `RootNode` such that you end up with an array of RootNodes where the first one is the main one. That way all async operations are parallel.
2. Then have a `mergeRootNodes` helper function that destructs `[head, ...tail]` and adds all programs inside `tail` to the additional programs of `head`.

lorisleiva (changes requested):
Thanks, I like the idea! I've just made a comment on the execution of this idea to make it more async.

packages/cli/src/parsedConfig.ts
plutohan: Done — conversions now run in parallel via Promise.all and a mergeRootNodes helper merges [head, ...tail]. Thanks!

packages/cli/src/parsedConfig.ts
```diff
@@ -46,13 +46,43 @@ async function parseConfig(
 ): Promise<ParsedConfig> {
     const idlPath = parseIdlPath(config, configPath, options);
     const idlContent = await importModuleItem({ identifier: 'IDL', from: idlPath });
-    const rootNode = await getRootNodeFromIdl(idlContent);
+    const rootNode = await mergeAdditionalIdls(idlContent, config.additionalIdls ?? [], configPath);
```
lorisleiva: Thanks, much better. I still think it would be nice if all RootNodes were recovered in the same way. Here the main RootNode has its own `importModuleItem` whereas the others at in the parallel execution. Do you know what I mean?

plutohan: Good point — the main IDL now goes through the same loadIdl helper as the others, so all root nodes are recovered identically inside the same Promise.all.

packages/cli/README.md
```diff
@@ -71,6 +71,7 @@ Here are the available options for the `run` command:
 The Codama configuration file defines an object containing the following fields:
 
 - `idl` (string): The path to the IDL file. This can be a Codama IDL or an Anchor IDL which will be automatically converted to a Codama IDL.
+- `additionalIdls` (array): An array of paths to additional Anchor IDLs. Their programs are included alongside the main IDL's program (as `additionalPrograms`), so a single config can generate clients for multiple programs in an Anchor workspace.
```
lorisleiva: nit: We should probably remove "Anchor" in the first sentence here as it implies Codama IDLs cannot be provided as additional IDLs. However I think it's a good idea to mention Anchor as an example use-case here since this is likely why users will reach for this option.

packages/cli/test/parsedConfig.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    return {
+        accounts: [],
+        constants: [],
+        definedTypes: [],
+        errors: [],
+        instructions: [],
+        kind: 'programNode',
+        name,
+        pdas: [],
+    } as unknown as ProgramNode;
```
lorisleiva: nit: Let's remove the type cast here so TS fails in the future when the nodes evolve.

packages/cli/test/parsedConfig.test.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function rootNodeWith(name: string): RootNode {
+    return {
+        additionalPrograms: [],
+        kind: 'rootNode',
+        program: programNode(name),
+        standard: 'codama',
+        version: '1.0.0',
+    } as unknown as RootNode;
```
lorisleiva: nit: same here

packages/cli/src/parsedConfig.ts
```diff
@@ -45,12 +45,34 @@ async function parseConfig(
...
+        ...additionalIdlPaths.map(path => loadIdl(path, 'additional IDL')),
+    ]);
+
+    const [mainIdl] = idls;
+    const rootNode = mergeRootNodes(idls.map(idl => idl.rootNode));
     const scripts = parseScripts(config.scripts ?? {}, configPath);
     const visitors = (config.before ?? []).map((v, i) => parseVisitorConfig(v, configPath, i, null));
 
-    return { configPath, idlContent, idlPath, rootNode, scripts, before: visitors };
+    return { configPath, idlContent: mainIdl.content, idlPath, rootNode, scripts, before: visitors };
```
lorisleiva, line 61: out-of-scope for this PR: I don't actually think `idlContent` is being used. Perhaps we should consider removing it altogether or making it match the config by adding a `additionalIdlContents`?

lorisleiva (approved):
Thank you! I've just made a few final nit comments but I'm happy with the implementation. 🍺

plutohan:
Addressed the nits: dropped `Anchor` from the `additionalIdls` description (kept the Anchor workspace example) and swapped the test node casts for the `programNode`/`rootNode` factories so future node changes fail the typecheck.

`idlContent` is indeed unused now. Rather than widen this PR, I'll open a follow-up to either remove it or add `additionalIdlContents` to match the config. Let me know which you'd prefer.

lorisleiva (approved):
Perfect, thank you!

PR #1021, opened 2026-07-13
docs: add Pinocchio source adapter

oxy-Op:
Adds a short documentation entry for native Pinocchio programs that want to produce Codama IDLs.

`pinocchio-idl` is a source-to-IDL adapter: it reads explicit Rust annotations from Pinocchio programs and emits Codama `rootNode` JSON. The generated file can then be used with existing Codama renderers.\n\nThis is docs-only and does not add a new package or renderer.


lorisleiva:
Hi there, sorry for security purposes we'd rather not link to external repositories as we cannot vouch for them. Also, have you seen this Codama project that extract a Codama IDL from any Rust code (with the ability to write custom plugins for advance usage / framework recognition)? https://github.com/codama-idl/codama-rs

oxy-Op:
Thanks for the context, that makes sense.

I’ll avoid linking external repos in Codama docs. I’ll take a look at codama-idl/codama-rs and see if there’s a better docs/example contribution.

PR #1025, opened 2026-07-20
Omit empty node arrays and type array attributes as optional

lorisleiva:
This PR makes every node array attribute skip-when-empty and default to `[]` when absent, implementing the array convention documented in `@codama/spec`. The node-types and nodes generators are updated so array attributes are typed as optional on the interfaces (e.g. `ProgramNode.accounts` is now `readonly accounts?: AccountNode[]`), their type-parameter constraints carry `| undefined`, and the factory functions omit an array from the built node when it is empty or absent, including when an explicit empty array is passed. The `node-types`, `nodes`, and `visitors-core` generated output is regenerated accordingly. Every downstream reader of a formerly-required array is guarded to normalise absent arrays to `[]`, so behaviour is unchanged at runtime and reading stays backwards-compatible with IDLs that still serialise empty arrays. Tests are updated for the slimmer output, new tests cover the omit-on-write and tolerate-on-read invariants, and array reads on factory-built nodes are tidied to reference the constructed nodes directly rather than re-reading them through the optionally-typed graph. The full build, lint, and test suites pass across all packages.

### DIFF
--- a/.changeset/big-pens-make.md
+++ b/.changeset/big-pens-make.md
@@ -1,0 +1,10 @@
+---
+'@codama/node-types': minor
+'@codama/nodes': minor
+---
+
+Omit empty array attributes from nodes, and default absent arrays to `[]` on read.
+
+Every node array attribute — whether previously required (e.g. `programNode.accounts`, `instructionNode.arguments`, `structTypeNode.fields`) or optional (e.g. `instructionNode.discriminators`) — is now omitted from the node when empty, and defaults to `[]` when absent. An absent array and an empty array are semantically identical, so this keeps encoded IDLs small (they are often uploaded on-chain) and makes adding or omitting an array attribute non-breaking. Reading remains backwards-compatible: IDLs that still serialise empty arrays continue to parse, and every visitor and accessor normalises an absent array to `[]`.
+
+**Breaking type change.** While runtime reads stay backwards-compatible, this is a breaking change at the type level for external consumers such as renderers and community packages. Array attributes are now typed as optional on the node interfaces (e.g. `ProgramNode.accounts` is now `readonly accounts?: AccountNode[]`), so any code that reads an array attribute without guarding — e.g. `program.accounts.map(…)` — must now coalesce with `?? []` (`(program.accounts ?? []).map(…)`). Type-parameter constraints also widen from `Array<T>` to `Array<T> | undefined`, so indexing into these types (e.g. `InstructionNode['arguments'][number]`) must be rewritten as `NonNullable<InstructionNode['arguments']>[number]`. Consumers should expect compile errors on the first unguarded array read after upgrading and add `?? []` guards accordingly.
--- a/packages/dynamic-address-resolution/src/codegen/collect-pda-nodes.ts
+++ b/packages/dynamic-address-resolution/src/codegen/collect-pda-nodes.ts
@@ -10,8 +10,8 @@ export function collectPdaNodesFromIdl(idl: RootNode): Map<string, PdaNode> {
         pdas.set(pda.name, pda);
     }
 
-    for (const ix of idl.program.instructions) {
-        for (const acc of ix.accounts) {
+    for (const ix of idl.program.instructions ?? []) {
+        for (const acc of ix.accounts ?? []) {
             if (!acc.defaultValue || acc.defaultValue.kind !== 'pdaValueNode') continue;
             const pdaDef = acc.defaultValue.pda;
             if (!pdaDef || pdaDef.kind !== 'pdaNode') continue;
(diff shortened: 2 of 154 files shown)


lorisleiva:
* **#1025** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1025?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1025?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Woody4618 (approved):

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

Makes every node array attribute skip-when-empty on write and tolerate absent on read, aligning the generated node types/factories with the array convention in `@codama/spec`. The generator changes are small and focused (four fragment files), and the regenerated `node-types`, `nodes`, and reader-side normalizations across `visitors-core`, `validators`, `dynamic-*`, and `nodes-from-anchor` all follow the same `?? []` pattern consistently. Tests cover the three important invariants: array attrs are omitted when empty, omitted even when an explicit `[]` is passed, and IDLs that entirely omit array attributes still parse and round-trip cleanly.

Overall the change is coherent and well-scoped. A few things worth surfacing:

## Semver / changeset

The changeset marks this as `minor` on `@codama/node-types` and `@codama/nodes`. Runtime reads stay backwards compatible (an IDL that still serialises `accounts: []` continues to parse), but at the **type level** this is a breaking change for any external consumer:

- `readonly accounts: AccountNode[]` → `readonly accounts?: AccountNode[]` on every array attribute means downstream renderers/visitors that do `program.accounts.map(…)` without `?? []` will fail to compile after upgrading.
- Type-parameter constraints widen from `Array<T>` to `Array<T> | undefined`, so consumers indexing into these types with e.g. `InstructionNode['arguments'][number]` will need to rewrite as `NonNullable<InstructionNode['arguments']>[number]` — exactly the pattern this PR had to apply to `encode-instruction-arguments.ts`.

That's not necessarily wrong for a `minor` under the repo's semver policy, but it's worth being explicit about in the changeset so external maintainers of renderers/visitors (`renderers-js`, `renderers-rust`, community packages, etc.) know they'll need to add `?? []` guards when they bump. Consider expanding the changeset body to call this out, or bumping to `major` if the repo's policy is stricter about surface-level type breaks.

## Correctness / consistency

The reader-side `?? []` sweep is thorough — I sampled `visitors-core`, `validators`, all `dynamic-*` packages, `nodes-from-anchor`, and the hand-written `nodes/src/*` wrappers and every array read is guarded. `isScalarEnum`'s `.every(…)` on an empty array returns `true` (vacuous truth), which matches the prior behaviour for an enum with zero variants and is fine.

One small nit in `packages/nodes/src/generated/RootNode.ts` — see inline.

- The generator diff is small (four files in `spec-generators/src/{nodes,nodeTypes}/fragments/`); worth reading first — everything else is either regenerated output or a mechanical `foo` → `foo ?? []` guard, and once you've internalised the rule the rest is easy to skim.
- Verify the `renderers-*` repos and any pinned community renderers still compile against these node types before releasing — the type change will surface as compile errors on the first `.accounts.map(…)` without `?? []`.
- No changes to `@codama/spec` itself in this PR — the convention it's aligning to must already be documented there; if not, that's a follow-up.

trevor-cortex:
Nit in `packages/nodes/src/generated/RootNode.ts`:

```ts
additionalPrograms: TAdditionalPrograms = [] as Array<ProgramNode> | undefined as TAdditionalPrograms,
```

The intermediate `| undefined` in the cast chain doesn't do anything useful — the runtime default is still `[]`, and the outer cast to `TAdditionalPrograms` already accepts an array. Was previously `[] as Array<ProgramNode> as TAdditionalPrograms`; simplest to drop the `| undefined` back out:

```ts
additionalPrograms: TAdditionalPrograms = [] as Array<ProgramNode> as TAdditionalPrograms,
```

Not a correctness issue — just noise in the generated output. If the generator template widened this deliberately to satisfy the new `| undefined` constraint, worth double-checking whether it's actually needed there.

trevor-cortex (approved):
Delta from last pass:

- **Generator fix**: `nodeFunctionPositionalArguments.ts` now special-cases array attributes so the intermediate cast widens through the bare element-array type (`Array<ItemNode>`) rather than `Array<ItemNode> | undefined`. Correct — widening an `[]` literal never needs the `| undefined`, and dropping it removes the noise I flagged. Test locks in both invariants (widens through bare array; no `| undefined` in output). ✅
- **`RootNode.ts` regenerated**: the cast chain is now `[] as Array<ProgramNode> as TAdditionalPrograms` — clean.
- **Bonus fix I hadn't flagged**: `additionalPrograms` in the body switched from an unconditional pass-through to the same `undefined && length > 0` conditional spread as every other array attribute. This closes a real inconsistency — the previous regenerated output was still emitting `additionalPrograms: []` for the empty case, which contradicted the PR's stated skip-when-empty invariant. Now `rootNode(programA)` produces a node with no `additionalPrograms` key, matching the rest.

Semver / changeset call-out from my prior review still stands as a separate discussion point, but nothing blocking. LGTM.

PR #1030, opened 2026-08-01
fix(nodes): uppercased snake case not correctly converted to snake case

ChiefWoods:
Anchor constant names are almost expected to come in SCREAMING_SNAKE_CASE due to Rust standards.

Updated `camelCase` to correctly handle the conversion.

### DIFF
--- a/packages/nodes/src/shared/stringCases.ts
+++ b/packages/nodes/src/shared/stringCases.ts
@@ -26,6 +26,11 @@ export function pascalCase(str: string): PascalCaseString {
 
 export function camelCase(str: string): CamelCaseString {
     if (str.length === 0) return str as CamelCaseString;
+    if (/^[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+$/.test(str)) {
+        return str
+            .toLowerCase()
+            .replace(/_([a-z0-9])/g, (_, character: string) => character.toUpperCase()) as CamelCaseString;
+    }
     const pascalStr = pascalCase(str);
     return (pascalStr.charAt(0).toLowerCase() + pascalStr.slice(1)) as CamelCaseString;
 }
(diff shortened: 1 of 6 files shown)


packages/nodes/src/shared/stringCases.ts
```diff
@@ -26,6 +26,11 @@ export function pascalCase(str: string): PascalCaseString {
 
 export function camelCase(str: string): CamelCaseString {
     if (str.length === 0) return str as CamelCaseString;
+    if (/^[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+$/.test(str)) {
+        return str
+            .toLowerCase()
+            .replace(/_([a-z0-9])/g, (_, character: string) => character.toUpperCase()) as CamelCaseString;
+    }
     const pascalStr = pascalCase(str);
```
lorisleiva: Thanks for the PR!

Since `camelCase` delegates to `pascalCase`, wouldn't it make more sense to integrate that logic to `pascalCase` or whichever casing helper is the lowest-common-denominator so it affects all casing that come from snake casing?

Also, whatever solution we apply here, would you mind also applying it to the `fragments` package?

https://github.com/codama-idl/codama/blob/main/packages/fragments/src/core/casing.ts

packages/nodes/src/shared/stringCases.ts
```diff
@@ -21,6 +21,11 @@ export function titleCase(str: string): TitleCaseString {
 }
 
 export function pascalCase(str: string): PascalCaseString {
+    if (/^[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+$/.test(str)) {
+        return str
+            .toLowerCase()
+            .replace(/(?:^|_)([a-z0-9])/g, (_, character: string) => character.toUpperCase()) as PascalCaseString;
+    }
     return titleCase(str).split(' ').join('') as PascalCaseString;
```
lorisleiva: Sorry looks like the lowest-common-denominator here is `titleCase` so might make more sense to move that there?

ChiefWoods: Apologies, wasn't immediately clear to me that `titleCase` is common denom.

lorisleiva (approved):
Thank you!

PR #1046, opened 2026-08-05
docs: fix broken package name and node doc links in README

latent-9:
A few links in the package READMEs point to names or paths that do not exist, so they 404 on GitHub.

- `packages/library/README.md`: the list links to `@codama/visitor-core` (`../visitor-core`), but the package is `@codama/visitors-core` (`packages/visitors-core/package.json`). The sibling entry already uses `@codama/visitors`.
- `packages/nodes/README.md`: three links end in `Nodemds` instead of `Node.md` (`ResolverValueNode`, `RemainderCountNode`, `SizeDiscriminatorNode`), and one drops the trailing `e` (`ZeroableOptionTypeNod` instead of `ZeroableOptionTypeNode`). All four target files exist with the corrected names.

Docs only, no code touched.


lorisleiva (approved):
Thanks!

PR #1056, opened 2026-08-05
docs: correct @codama/renderers-js package name in CLI README

latent-9:
The CLI README refers to the JavaScript renderer as `@codama/renderer-js`, but the package is `@codama/renderers-js` (plural) — which is also the repo the link already points to (`codama-idl/renderers-js`). This corrects the package name in the link text.


lorisleiva (approved):

latent-9:
sankyou

PR #1081, opened 2026-08-18
Migrate from eslint and prettier to oxlint and oxfmt

lorisleiva:
This PR migrates the monorepo's linting and formatting to the oxc toolchain, following the pattern established in kit-plugins and the spec repo: a single root oxlint.config.ts and oxfmt.config.ts extending @solana-config/oxc replace the eslint flat config, the prettier config, and sixteen .prettierignore files, and linting now runs once from the root instead of as seventeen turbo tasks. Both configs share an identical ignore list, including a repo-wide TOML/Cargo.lock exclusion since the anchor test fixture is a Rust project whose files are hash-pinned build inputs of the committed program dumps. Type-aware rules run through oxlint-tsgolint, which required moving the root tsconfig from the removed node10 resolution to Bundler with module Preserve, giving dynamic-client noEmit since its tests import loose .mjs build scripts, and replacing the blanket lib: [] in ten packages with lib: ["ES2022"] so typescript-go resolves built-ins instead of error-typing them. That last change made type-aware analysis meaningful in those packages and surfaced genuine latent bugs which this PR fixes: three error contexts in @codama/errors declared types their runtime values could never satisfy (readonly validation items, the dangling dependency reference on invalid default-value dependencies), dynamic-codecs passed a link node where the resolved enum type was declared, two promises went unhandled, and the anchor sync script logged [object Object]. Existing rule exemptions are preserved (sort-keys stays off for cli, nodes and node-types; order-sensitive fixtures keep their inline disables under the renamed rule), and two broken autofixes were caught and reverted during migration. Formatting now covers root-level files and package manifests, which were previously unchecked in CI, and dependabot gains an oxc group.


lorisleiva:
* **#1081** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1081?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1081?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

aleksola (approved):

PR #1089, opened 2026-08-20
Restore formatting and linting of generated files

lorisleiva:
This PR repairs the post-generation lint step of the spec-generation pipeline, which has been silently broken since the oxlint/oxfmt migration (#1081). That migration removed the package-level `lint:fix` scripts, so the `pnpm --filter ... lint:fix` step in `spec-generators`' `generate` script matched no scripts and no-oped, and it also ignored `**/generated/` in both oxc configs, so freshly generated output would land completely unformatted and unlinted. The `generate` script now runs the root `pnpm -w lint:fix`, and `**/generated/` is removed from both ignore lists, which stay in sync; rules that are inappropriate for machine-written output are instead disabled via a new oxlint override for `**/generated/**` (currently just `sort-keys`, which is auto-fixable and could otherwise silently reorder deliberately-ordered generated structures). Generated files were linted and formatted throughout the eslint/prettier era, so this is a return to form; oxlint reports zero violations on them under the full rule set. The committed generated files are reformatted accordingly, a 44-file whitespace-only delta. Verified by regenerating against the pinned `@codama/spec@1.8.0`: the pipeline now reproduces the committed files exactly and is idempotent, confirming there was no semantic drift.

### DIFF
--- a/oxfmt.config.ts
+++ b/oxfmt.config.ts
@@ -4,7 +4,6 @@ const solanaFmt = require('@solana-config/oxc/oxfmt');
 // Keep in sync with oxlint.config.ts.
 const ignorePatterns = [
     '**/dist/',
-    '**/generated/',
     '**/target/',
     '**/test-ledger/',
     '**/idls/',
--- a/oxlint.config.ts
+++ b/oxlint.config.ts
@@ -4,7 +4,6 @@ const solanaConfig = require('@solana-config/oxc/oxlint');
 // Keep in sync with oxfmt.config.ts.
 const ignorePatterns = [
     '**/dist/',
-    '**/generated/',
     '**/target/',
     '**/test-ledger/',
     '**/idls/',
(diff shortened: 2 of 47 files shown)


lorisleiva:
* **#1091** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1091?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1090** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1090?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1089** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1089?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1089?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1095, opened 2026-08-20
Add @codama/upgrade

lorisleiva:
This PR adds the `@codama/upgrade` package: an append-only chain of pure, hand-written functions, each upgrading a Codama document exactly one major to the next. `upgrade()` detects the source major from the document's `version`, refuses pre-1.0 or unparsable versions with the new `CODAMA_ERROR__UNSUPPORTED_VERSION` code, refuses future majors with the existing `CODAMA_ERROR__VERSION_MISMATCH` (whose message now covers both directions), runs the applicable upgrade functions as sequential `if (major <= N)` blocks, and restamps the result with the latest `CODAMA_VERSION`. The chain is empty while the latest spec major is 1, so upgrading is identity plus restamp; adding a future major means one new function and one new `if` block, and because the functions and frozen types are committed source with zero dependencies, every major back to 1.0.0 stays upgradable forever. The frozen v1 node types are generated as part of the default `generate` orchestrator from a new `@codama/spec-v1` aliased exact pin in `spec-generators`, so the CI freshness check guards the snapshot automatically and it only changes when that pin or the generator code deliberately changes; after the v2 transition, the living pin moves to 2.x while the alias keeps pointing at the last 1.x. The snapshot is committed under `src/v1` alongside frozen copies of the hand-written `brands`/`Docs`/`Version` siblings, is compile-time only, and is exposed as the type-only `v1` namespace, adding zero runtime bytes. `upgradeFromJson` covers the JSON path and `upgradeToLatestVisitor` provides the CLI-friendly preprocessing boundary; it is also the package's default export, so the bare `"@codama/upgrade"` module name works as a `before` visitor in CLI configs. Tests cover detection, restamping, content preservation, all error paths, a compile-time guarantee that the frozen `v1` types describe current v1 documents, and upgraded fixtures passing `@codama/validators` as a smoke check. Upgrade functions are pure JSON-tree-in, JSON-tree-out transforms with no environment access, keeping the chain portable to `codama-rs`. This is Phase 2 of the v2 migration framework plan.


lorisleiva:
* **#1095** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1095?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

trevor-cortex (approved):
## Summary

This PR introduces `@codama/upgrade`, Phase 2 of the v2 migration framework. The core is a small, dependency-light runtime (`upgrade`, `upgradeFromJson`, `upgradeToLatestVisitor`) built around an append-only chain of pure single-major upgrade functions — currently empty while the latest spec major is 1, so upgrading is identity + restamp. The bulk of the diff is the frozen v1 node-type snapshot under `src/v1`, generated from a new `@codama/spec-v1` aliased exact pin in `spec-generators` and exposed as the type-only `v1` namespace. Two error changes in `@codama/errors` support the version gating (`CODAMA_ERROR__UNSUPPORTED_VERSION` new, `CODAMA_ERROR__VERSION_MISMATCH` message extended), each covered by an appropriately-leveled changeset.

The design is sound and the implementation is clean: the freeze pipeline being wired into the default `generate` orchestrator means the CI freshness check guards the snapshot for free, the exact `npm:@codama/spec@1.9.1` alias makes the pin's intent explicit, and keeping the upgrade functions as pure JSON-tree transforms keeps the chain portable to `codama-rs`. Tests cover the error paths, restamping, content preservation, the compile-time v1-shape guarantee, and a validators smoke check.

## Things to watch out for

- **Same-major, newer-minor documents are silently restamped down** (inline comment on `upgrade.ts`). A 1.12 document processed by tooling pinned to 1.9 passes the major check and gets restamped to 1.9.1, even though it may carry node kinds the installed packages don't know about. Probably a deliberate "minors are additive, best-effort" stance, but worth confirming it's intentional since the restamp erases the evidence.
- **`Object.freeze` on the result is shallow** — only the top-level root is frozen. The test name ("it returns a frozen document") slightly oversells this. Fine if the intent is just to freeze the restamped wrapper, but worth knowing.
- **`upgradeFromJson` lets `JSON.parse` errors escape raw** (inline comment) — malformed JSON surfaces as a bare `SyntaxError` rather than a `CodamaError`.

- Everything under `packages/upgrade/src/v1/generated/` is machine-generated from the pinned spec and CI-guarded; it can be skimmed rather than line-reviewed. The hand-written frozen siblings (`brands.ts`, `Docs.ts`, `Version.ts`) are faithful copies of their `@codama/node-types` counterparts and carry "do not edit" headers.
- `packages/upgrade/package.json` (exports map, scripts, browserslist) mirrors the sibling packages' layout — worth a quick diff against e.g. `packages/errors/package.json` if anything looks off, but nothing jumped out.
- The `@codama/upgrade` changeset is `major` on a `0.0.0` package, which releases it at 1.0.0 — intentional for a first release.
- `spec-generators` is internal/unpublished, so no changeset is needed for its changes.

PR #1096, opened 2026-08-21
Add checkCodamaVersionVisitor and document the CLI upgrade pattern

lorisleiva:
This PR adds a lightweight `checkCodamaVersionVisitor` to `@codama/visitors`, filling the gap where the CLI accepts any Codama-standard JSON verbatim with zero version checking. The visitor mirrors the check performed by `createFromRoot`: it compares the visited IDL's major against the supported `CODAMA_VERSION` via `getCodamaVersionMajor` and throws the existing `CODAMA_ERROR__VERSION_MISMATCH` on mismatch or unparsable versions, returning the document unchanged otherwise. Following the no-implicit-auto-upgrade policy, nothing is wired into the CLI automatically; instead, the CLI README gains two recipes documenting the `before` visitor pattern: upgrading legacy IDLs via `@codama/upgrade` and failing fast on incompatible ones via `@codama/visitors#checkCodamaVersionVisitor`. The visitors README documents the new visitor alongside the existing ones. This is Phase 3 of the v2 migration framework plan.

### DIFF
--- a/packages/visitors/test/checkCodamaVersionVisitor.test.ts
+++ b/packages/visitors/test/checkCodamaVersionVisitor.test.ts
@@ -1,0 +1,46 @@
+import { CODAMA_ERROR__VERSION_MISMATCH, CodamaError } from '@codama/errors';
+import { CODAMA_VERSION, CodamaVersion, programNode, rootNode } from '@codama/nodes';
+import { visit } from '@codama/visitors-core';
+import { expect, test } from 'vitest';
+
+import { checkCodamaVersionVisitor } from '../src';
+
+test('it accepts IDLs sharing the spec major and returns them unchanged', () => {
+    // Given IDLs whose versions share the spec major.
+    const program = programNode({ name: 'myProgram', publicKey: '1111' });
+    const latest = rootNode(program);
+    const oldest = { ...rootNode(program), version: '1.0.0' as CodamaVersion };
+
+    // When we visit them with the check visitor, then they are returned unchanged.
+    expect(visit(latest, checkCodamaVersionVisitor())).toBe(latest);
+    expect(visit(oldest, checkCodamaVersionVisitor())).toBe(oldest);
+});
+
+test('it rejects IDLs from another spec major', () => {
+    // Given IDLs whose versions belong to other spec majors.
+    const program = programNode({ name: 'myProgram', publicKey: '1111' });
+    const newer = { ...rootNode(program), version: '99.0.0' as CodamaVersion };
+    const older = { ...rootNode(program), version: '0.21.3' as CodamaVersion };
+
+    // When we visit them with the check visitor, then we expect version mismatch errors.
+    expect(() => visit(newer, checkCodamaVersionVisitor())).toThrow(
+        new CodamaError(CODAMA_ERROR__VERSION_MISMATCH, { codamaVersion: CODAMA_VERSION, rootVersion: '99.0.0' }),
+    );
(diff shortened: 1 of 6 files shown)


lorisleiva:
* **#1096** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1096?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1096?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/visitors/test/checkCodamaVersionVisitor.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import { CODAMA_VERSION, CodamaVersion, programNode, rootNode } from '@codama/nodes';
+import { visit } from '@codama/visitors-core';
+import { expect, test } from 'vitest';
+
+import { checkCodamaVersionVisitor } from '../src';
+
+test('it accepts documents sharing the spec major and returns them unchanged', () => {
+    // Given documents whose versions share the spec major.
+    const program = programNode({ name: 'myProgram', publicKey: '1111' });
+    const latest = rootNode(program);
```
trevor-cortex, line 11: Nit: this hardcodes major `1`, so the "accepts documents sharing the spec major" test silently becomes a wrong-major case whenever the spec moves to `2.x`. Deriving it keeps the test self-maintaining:

```ts
const oldest = { ...rootNode(program), version: `${getCodamaVersionMajor(CODAMA_VERSION)}.0.0` as CodamaVersion };
```

(Same reasoning applies to `0.21.3`/`99.0.0` in the rejection test, but those are far less likely to ever collide with the real major.)

trevor-cortex (approved):
## Summary

This PR adds `checkCodamaVersionVisitor` to `@codama/visitors`: a small root-node visitor that mirrors the version check `createFromRoot` performs (`validateCodamaVersion` in `packages/library/src/codama.ts`) — comparing the document's major against `CODAMA_VERSION` via `getCodamaVersionMajor` and throwing `CODAMA_ERROR__VERSION_MISMATCH` on mismatch or unparsable versions. Nothing is wired into the CLI automatically, consistent with the no-implicit-auto-upgrade policy; instead the CLI README gains two `before`-visitor recipes (fail-fast check, and upgrade via `@codama/upgrade`), and the visitors README documents the new visitor.

The implementation is faithful to `validateCodamaVersion` (I diffed the two — same null-handling, same error payload shape), the export is correctly alphabetized in `src/index.ts`, both README recipe sections land in the correct alphabetical position, and the changeset is a `minor` bump on `@codama/visitors`, which matches precedent for new visitor additions. Tests cover the three meaningful branches: same-major acceptance (including identity of the returned node), cross-major rejection in both directions, and unparsable versions.

## Notes for reviewers

- **Intentional duplication with `validateCodamaVersion`**: the check logic now lives in two places. Delegating isn't straightforward — `@codama/visitors` can't depend on `@codama/library` without a cycle — so the mirror is a reasonable call. If a third caller appears, a shared assert next to `getCodamaVersionMajor` in `@codama/nodes` would be the natural home; not worth blocking on now.
- **Recipe interplay**: the visitors README correctly points users toward `@codama/upgrade` as the "upgrade instead of reject" alternative. If both are ever used together, ordering in `before` matters (upgrade first, then check), but the docs don't suggest combining them, so no change needed.
- One minor test-fragility nit inline.

PR #1126, opened 2026-08-28
Fix byte-array PDA seeds deriving with a Borsh length prefix

lorisleiva:
This PR resurrects #1101 by @latent-9, which I accidentally closed whilst cleaning up stale branches (sorry again! 🙏). The original commit is cherry-picked verbatim from the PR ref, so authorship is preserved and it already has an approving review on the original thread.

In short: a `Vec<u8>` instruction argument used as a PDA seed kept its Borsh `sizePrefixTypeNode(u32, bytesTypeNode())` type when converted from an Anchor IDL, whereas Anchor derives byte-array seeds from the raw bytes. This meant generated clients computed a different PDA than the on-chain program. The fix mirrors the existing `isBorshString` special case with an `isBorshBytes` counterpart that emits a raw `bytesTypeNode()` seed.

PR #1127, opened 2026-08-29
Fix variable string and bytes PDA seeds in legacy Anchor IDLs

lorisleiva:
This PR is a follow-up to #1126, which fixed byte-array PDA seeds for Anchor v01 IDLs. The legacy v00 path has the same bug: `pdaNodeFromAnchorV00` strips the Borsh size prefix from constant `string` and `bytes` seeds but not from variable ones, so a variable seed derives a different PDA than the on-chain program. This change hoists the seed type conversion into a single helper used by both branches, so constant and variable seeds are now consistent by construction.

### DIFF
--- a/.changeset/long-stamps-bow.md
+++ b/.changeset/long-stamps-bow.md
@@ -1,0 +1,5 @@
+---
+'@codama/nodes-from-anchor': patch
+---
+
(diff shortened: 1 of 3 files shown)


lorisleiva:
* **#1127** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1127?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1127?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1149, opened 2026-09-11
chore: minor security improvements

mikhd:
### Summary

  - `.npmrc` - pin resolution to `https://registry.npmjs.org/`.
  - `pnpm-workspace.yaml` - add `onlyBuiltDependencies: []`. (pnpm 10 already blocks dependency install and build scripts by default).
  - Workflows - pin the third-party actions to full commit SHAs.
    - `dependabot.yml` - add the `github-actions` so the new SHA pins do not rot.
  - `@codama/release-tools` - bumped `v1.0.0` -> `v1.2.0` to match the workflow pin.
  - `anchor:build` - add `--locked`.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -44,8 +44,8 @@ updates:
       github-actions:
         patterns:
           - "*"
-        # Minor and patch bumps are grouped into one PR;
-        # majors are split out so each can be reviewed on its own.
+        # Minor and patch bumps are grouped into one PR.
+        # Majors are split out so each can be reviewed on its own.
         update-types:
           - "minor"
           - "patch"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -34,8 +34,8 @@ updates:
 
   # GitHub Actions used across .github/workflows/*.
   # Third-party actions are pinned to commit SHAs and this keeps the pins from rotting.
-  - package-ecosystem: "github-actions"
-    directory: "/"
+  - package-ecosystem: 'github-actions'
+    directory: '/'
     schedule:
       interval: weekly
     cooldown:
@@ -43,9 +43,9 @@ updates:
     groups:
       github-actions:
         patterns:
-          - "*"
+          - '*'
         # Minor and patch bumps are grouped into one PR.
         # Majors are split out so each can be reviewed on its own.
         update-types:
-          - "minor"
-          - "patch"
+          - 'minor'
+          - 'patch'
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -39,7 +39,7 @@ updates:
     schedule:
       interval: weekly
     cooldown:
-      default-days: 14
+      default-days: 7
     groups:
       github-actions:
         patterns:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,5 @@
+# Dependabot version updates. Opens PRs to keep dependency versions fresh.
+# Docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 version: 2
 updates:
   - package-ecosystem: 'npm'
@@ -29,3 +31,21 @@ updates:
           - 'oxfmt'
           - 'oxlint-tsgolint'
           - '@solana-config/oxc'
+
+  # GitHub Actions used across .github/workflows/*.
+  # Third-party actions are pinned to commit SHAs and this keeps the pins from rotting.
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: weekly
+    cooldown:
+      default-days: 14
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+        # Minor and patch bumps are grouped into one PR;
+        # majors are split out so each can be reviewed on its own.
+        update-types:
+          - "minor"
+          - "patch"
--- a/.github/workflows/cut.yml
+++ b/.github/workflows/cut.yml
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   cut:
-    uses: codama-idl/release-tools/.github/workflows/cut.yml@v1
+    uses: codama-idl/release-tools/.github/workflows/cut.yml@cc0b8c9589e43c869233e2533dacb35f25b188e4 # v1.2.0
     with:
       tracking-issue: ${{ inputs.tracking-issue }}
       announce: ${{ inputs.announce }}
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,3 @@
-# Pin resolution to the official registry so an inherited global or org config cannot redirect it.
+# Default registry for package-name lookups. Scope-specific entries in an inherited user or global
+# .npmrc still take precedence, so the lockfile pins stay the real integrity control.
 registry=https://registry.npmjs.org/
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,2 @@
+# Pin resolution to the official registry so an inherited global or org config cannot redirect it.
+registry=https://registry.npmjs.org/
--- a/packages/dynamic-client/package.json
+++ b/packages/dynamic-client/package.json
@@ -53,7 +53,7 @@
         "build": "rimraf dist && tsup && tsc -p ./tsconfig.declarations.json",
         "dev": "vitest --project node",
         "test": "pnpm generate-program-types && pnpm test:types && pnpm test:treeshakability && pnpm test:unit",
-        "anchor:build": "cd test/programs/anchor && anchor build --ignore-keys",
+        "anchor:build": "cd test/programs/anchor && anchor build --ignore-keys -- --locked",
         "anchor:sync:build": "pnpm anchor:build && node ./scripts/anchor/sync-anchor-builds.mjs && pnpm anchor:generate-idl && pnpm generate-program-types",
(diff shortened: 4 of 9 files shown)


.npmrc
```diff
@@ -0,0 +1,2 @@
+# Pin resolution to the official registry so an inherited global or org config cannot redirect it.
+registry=https://registry.npmjs.org/
```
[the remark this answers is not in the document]
mikhd: https://github.com/codama-idl/codama/pull/1149/commits/0c065faa835aa70b69a270973780667250bdb9fd
Slightly overstated, but right that this only covers the default registry - comment narrowed to say so. Not maintaining a scope list: 23 scopes, mostly transitive, and a new one would be silently unprotected. Integrity hashes in the lockfile plus --frozen-lockfile are the actual control.

pnpm-workspace.yaml
```diff
@@ -1,2 +1,5 @@
 packages:
   - "packages/*"
+
+# No dependency may run install or build scripts.
```
rogaldh, line 4:
```suggestion
minimumReleaseAge: 10080
# No dependency may run install or build scripts.
```

I'd suggest adding a delay for the dependencies to sync with the delay for dependabot.
But this requires updating to pnpm@10.16

rogaldh (approved):

lorisleiva:
@trevor-cortex

trevor-cortex (commented):
## Summary

Supply-chain hardening, no runtime code changes: pins the default npm registry via a new `.npmrc`, makes the pnpm "no dependency build scripts" policy explicit with `onlyBuiltDependencies: []`, pins all third-party actions and the `release-tools` reusable workflows to full commit SHAs (with version comments), adds a `github-actions` Dependabot ecosystem so those pins keep moving, bumps `@codama/release-tools` from `v1.0.0` to `v1.2.0`, and adds `--locked` to the Anchor test-program build.

All sensible. A few things worth a second look before merging:

## Things to watch out for

1. **Two different SHAs are labelled `v1.2.0` for `release-tools`.** The workflows pin `cc0b8c9589e43c869233e2533dacb35f25b188e4 # v1.2.0`, while `pnpm-lock.yaml` resolved `github:codama-idl/release-tools#v1.2.0` to `97d4c3bdd1bf5623cdc53a083bf2a00bf24843c5`. pnpm resolves through the GitHub API to the peeled commit, so if the tag is annotated the workflow value might be the tag object rather than the commit (which `uses:` won't accept), or the two were taken from different refs. I couldn't reach the `release-tools` repo to check which one is right — a quick `git ls-remote https://github.com/codama-idl/release-tools v1.2.0` would settle it. (Detail inline on `cut.yml`; `promote.yml` L10 has the same pin.)

2. **Dependabot auto-merge now covers action bumps.** The existing `dependabot` job in `main.yml` auto-approves and auto-merges any Dependabot PR whose `update-type` is minor or patch. With the new `github-actions` ecosystem grouping minor/patch bumps into one PR, SHA updates to `pnpm/action-setup`, `changesets/action`, `dependabot/fetch-metadata` and the `release-tools` reusable workflows will land without a human looking at them. The 7-day cooldown mitigates the "compromised tag" scenario a lot, but it's worth making sure that's a deliberate trade-off — the whole point of SHA pinning is that a human decides when the SHA moves. If you'd rather keep eyes on those, the job could skip when `steps.metadata.outputs.package-ecosystem == 'github_actions'`. (Inline on `dependabot.yml`.)

3. **First-party actions are left on tags.** `actions/checkout@v7`, `actions/setup-node@v7` and `actions/create-github-app-token@v3` (unchanged lines in `main.yml`) are still floating. That's a common and defensible line to draw for `actions/*`, but `create-github-app-token` is the step that handles the release app's private key, so it's arguably the one most worth pinning. Dependabot would maintain it the same way as the others. Not blocking — just flagging the inconsistency so it's a choice rather than an omission.

4. **`anchor:build --locked` isn't exercised in CI.** Nothing in `main.yml` runs the Anchor build, so this change is only validated by whoever runs `anchor:sync:build` locally. `anchor build -- <args>` forwards to `cargo build-sbf`, which should pass `--locked` through to cargo, but worth confirming it actually took effect (e.g. it fails if you tweak a version in `Cargo.toml` without updating the lock). `test/programs/anchor/Cargo.lock` is committed, so the flag has something to lock against.

5. **`onlyBuiltDependencies: []`** — as the description says, this is pnpm 10's default behaviour, so the change is declarative. One thing to check: pnpm 10 prints an "Ignored build scripts: … run `pnpm approve-builds`" warning after install for packages that wanted to run scripts (esbuild etc.). If an empty allowlist doesn't suppress it, `ignoredBuiltDependencies` is the idiomatic way to record the per-package decision and silence the nag.

- I could not independently verify that the `pnpm/action-setup`, `changesets/action` and `dependabot/fetch-metadata` SHAs match their `# vX.Y.Z` comments. If someone has `gh` handy, `gh api repos/<owner>/<repo>/git/ref/tags/<tag>` for each is a 30-second check and the most valuable thing to do on this PR.
- `.npmrc`'s scoped-registry caveat (raised by Copilot) is already acknowledged in the file comment; the lockfile integrity hashes remain the real control. Nothing further needed there.
- The lockfile still reports `version: 1.0.0` for the `release-tools` tarball — that's just `release-tools`' own `package.json` not being bumped on tag, not a problem for this repo, but it's why the tag→SHA mapping in point 1 deserves a look rather than being assumed.

.github/workflows/cut.yml
```diff
@@ -18,7 +18,7 @@ permissions:
 
 jobs:
   cut:
-    uses: codama-idl/release-tools/.github/workflows/cut.yml@v1
+    uses: codama-idl/release-tools/.github/workflows/cut.yml@cc0b8c9589e43c869233e2533dacb35f25b188e4 # v1.2.0
```
trevor-cortex: This pins `v1.2.0` to `cc0b8c95…`, but `pnpm-lock.yaml` in the same PR resolved `github:codama-idl/release-tools#v1.2.0` to `97d4c3bd…`. Both can't be the commit the tag points at. pnpm resolves via the GitHub API to the peeled commit SHA, so if `v1.2.0` is an annotated tag, `cc0b8c95…` may be the tag object — which `uses:` won't resolve. Could you double-check with `git ls-remote https://github.com/codama-idl/release-tools v1.2.0` (it prints both the tag object and the `^{}` peeled commit) and align both pins to the commit? Same value is used in `promote.yml` L10.

.github/dependabot.yml
```diff
@@ -29,3 +31,21 @@ updates:
...
+      default-days: 7
+    groups:
+      github-actions:
+        patterns:
+          - '*'
+        # Minor and patch bumps are grouped into one PR.
+        # Majors are split out so each can be reviewed on its own.
+        update-types:
+          - 'minor'
+          - 'patch'
```
trevor-cortex, line 51: Heads-up on an interaction with the existing `dependabot` job in `main.yml`: it auto-approves and auto-merges any Dependabot PR whose `update-type` is `semver-minor` or `semver-patch`, regardless of ecosystem. With this block, the grouped minor/patch action bumps (including the `release-tools` reusable-workflow pins that run with the release app token and `NPM_TOKEN`) will merge without a human looking at the new SHA.

The 7-day cooldown helps a lot against a compromised-tag scenario, so this may well be the trade-off you want — just making sure it's deliberate. If you'd rather keep human eyes on action pins, the auto-approve/auto-merge steps could add `&& steps.metadata.outputs.package-ecosystem != 'github_actions'`.

packages/dynamic-client/package.json
```diff
@@ -53,7 +53,7 @@
         "build": "rimraf dist && tsup && tsc -p ./tsconfig.declarations.json",
         "dev": "vitest --project node",
         "test": "pnpm generate-program-types && pnpm test:types && pnpm test:treeshakability && pnpm test:unit",
-        "anchor:build": "cd test/programs/anchor && anchor build --ignore-keys",
+        "anchor:build": "cd test/programs/anchor && anchor build --ignore-keys -- --locked",
```
trevor-cortex, line 56: Nothing in CI runs `anchor:build`, so this is only validated locally. `anchor build -- <args>` forwards to `cargo build-sbf`, which should pass `--locked` through to cargo — worth a quick sanity check that it actually errors when `Cargo.lock` is stale (e.g. bump a dep version in `Cargo.toml` without updating the lock) so we know the flag isn't being silently swallowed.

.github/dependabot.yml
lorisleiva: I think that's fine.

.github/workflows/cut.yml
lorisleiva: That's a very good point. Also the whole point of this homemade action is to sync the various settings and workflows of the Codama repos so pinning this would break this as we'd need to manually bump all of them before they can be sync'ed. Since this is a repo _we_ control, I'd say let's keep the `v1` tag.

@mikhd Please revert this. 🙏

packages/dynamic-client/package.json
lorisleiva: I think that's fine.

.github/workflows/cut.yml
mikhd: @trevor-cortex @lorisleiva  Thank you! Updated https://github.com/codama-idl/codama/pull/1149/commits/9957e193606a191dff35960e3478ccc03f3bc520

pnpm-workspace.yaml
mikhd: Yeah, i deliberately decided not to add this one in current changes.

lorisleiva (approved):
Thanks!

PR #1158, opened 2026-09-17
[2.x] Regenerate the node types from the v2 spec

lorisleiva:
This PR repoints `@codama-internal/spec-generators` at `@codama/spec@2.0.0-rc.0` and regenerates `@codama/node-types`, `@codama/nodes` and `@codama/visitors-core` against the v2 spec. The hand-written packages are adapted to the v2 shapes in the follow-up PRs, so they remain red until those land.

## Generator changes

- **Single-major generator.** Drops the `@codama/spec-v1` alias and the frozen v1 generation step. The v1 node types in `@codama/upgrade` are now a frozen static snapshot (`packages/upgrade/src/v1/README.md`); the generator only renders the spec on its own branch, and prior majors are ported forward by patch.
- **Base attributes.** Appends `spec.base` (the universal `plugins`) to every node, so `plugins` is a `TPlugins` type parameter on every interface and constructor and is visited by the identity/merge visitors. `textNode` gains `plugins` too.
- **`docs` === `text`.** Both render as `string | TextNode` and are visited (guarding the string-or-`textNode` duality). Deletes the v1 `Docs` alias and the `parseDocs`/`DocsInput` array layer.
- **Identifier casing preserved.** Constructors brand identifiers via `identifierString` rather than re-casing them, per v2 dropping the camelCase mandate.
- **`CODAMA_VERSION` names the spec shape.** Pre-release metadata is stripped (`2.0.0-rc.0` → `2.0.0`); the `CodamaVersion` type stays strict.
- **Constructor ergonomics.** Trailing optional positional parameters fold into the `options` bag (e.g. `constantNode(identifier, type, value, { docs })`); `pluginNode(namespace, payload?, { plugins })` keeps its payload positional.
- **Branded-string validation.** Factories accept plain `string` for identifier/namespace/path/integer/decimal attributes and validate them through `identifierString`/`namespaceString`/`pathString`/`integerString`/`decimalString`, which throw `CODAMA_ERROR__INVALID_BRANDED_STRING` on a malformed value.
- **Dead meta-model removed.** Nested-union / tuple / float handling deleted.

## Verification

Generated output type-checks, lints and is idempotent; the frozen v1 snapshot is untouched. The v2 changes are documented in a single major changeset alongside the seeded one from the cut.


lorisleiva:
* **#1171** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1162** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1162?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1158** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1158?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1158?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #1162, opened 2026-09-18
[2.x] Rename node name attribute to identifier

lorisleiva:
This PR renames the v2 `name` node attribute to `identifier` across all hand-written package source — a 1:1 rename with no behavioural changes.

## What changed

- `.name` member accesses on renamed node kinds → `.identifier` (programNode, accountNode, instructionNode, errorNode, eventNode, pdaNode, constantNode, definedTypeNode, instructionAccountNode, instructionRemainingAccountsNode, providedNode, structFieldTypeNode, enumVariantTypeNode, variablePdaSeedNode, the value nodes, and all link nodes).
- `name:` object-literal keys constructing those nodes → `identifier:`.
- `update*Visitor` config reads (`updates.name`, `argUpdate.name`) → `identifier`, since those config types are `Partial<…NodeInput>` which use `identifier` in v2.
- The `'name' in node` narrowing sites — which type-check against the v2 `Node` union but are runtime no-ops (`NodeSelector.ts`, `NodePath.ts`, `getDebugStringVisitor.ts` default branch) — switched to `'identifier' in node` / `node.identifier`, plus the selector example in the README.

## Deliberately left untouched (not the `name` attribute)

- Raw Anchor IDL reads (`idl.*.name`, `'name' in field`), commander `.name()`, error-context `name:` keys, projected plain-object keys, and anything inside removed-node guards (`argumentValueNode`, `resolverValueNode`, `instructionArgumentLinkNode`) — handled when each package is adapted to v2.
- `NodeSelector`'s `camelCase(name) !== node.identifier` comparison: v2 identifiers are no longer camelCase-normalised, so the comparison semantics need a case-fold rethink. Cast to `string` here to keep the rename compiling; the semantic fix is a selector follow-up.

## Note on removed nodes

`instructionArgumentNode` is a removed node in v2, but its construction sites still had `name:` → `identifier:` renamed here (harmless — they become struct fields, which use `identifier`). Those sites are not fully migrated; the argument→data rework happens when each package is adapted. Similarly `fieldDiscriminatorNode.name` → `path` and `accountFieldValueNode` → `accountDataValueNode` are out-of-scope shape changes left for those PRs.

## Status

Mechanical rename only; packages are not yet green (numeric/transform/removed-node work follows per package). Every remaining `'name'` error in `src/` sits inside a removed-node branch. `test/` and generated files are untouched.


lorisleiva:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/codama-idl/codama/1162?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1171** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1162** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1162?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1162?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1158** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1158?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/visitors-core/src/getDebugStringVisitor.ts
```diff
@@ -64,17 +64,17 @@ function getNodeDetails(node: Node): string[] {
...
+            return [node.identifier];
         case 'accountLinkNode':
         case 'definedTypeLinkNode':
         case 'instructionAccountLinkNode':
         case 'instructionArgumentLinkNode':
         case 'instructionLinkNode':
         case 'pdaLinkNode':
         case 'programLinkNode':
-            return [node.name];
+            return [node.identifier];
```
trevor-cortex, line 77: Two things in this function:

- **nit:** `instructionArgumentLinkNode` is grouped into this `node.identifier` branch, whereas `LinkableDictionary.ts` deliberately leaves the `instructionArgumentLinkNode` branches on `.name` as a removed-node guard. Both are dead in v2 so it doesn't matter functionally, just inconsistent with the rule in the description.
- **actual miss** (anchoring here; the real line is the `default:` branch at ~L115): `return 'name' in node ? [node.name] : [];` still uses `name`. Because of the `in` narrowing it compiles fine against v2 types, but at runtime no v2 node has `name`, so the fallback silently prints nothing for `accountNode`, `instructionNode`, `definedTypeNode`, `pdaNode`, struct fields, enum variants, `variablePdaSeedNode`, the value nodes, etc. Should be `'identifier' in node ? [node.identifier] : []`.

trevor-cortex (changes requested):
## Summary

Mechanical `name` → `identifier` rename across hand-written `src/` in the dynamic-*, nodes-from-anchor, validators, visitors-core and visitors packages. I went through all 87 files: every changed line is a faithful 1:1 swap, the `update*Visitor` config reads (`updates.identifier`, `argUpdate.identifier`) line up with the `Partial<…NodeInput>` types, and the value-node/link-node accesses I spot-checked (`accountValueNode`, `accountBumpValueNode`, `pdaSeedValueNode`, link nodes) all have `identifier: IdentifierString` in the regenerated `@codama/node-types`. Removed-node branches (`argumentValueNode`, `resolverValueNode`, `instructionArgumentLinkNode`) are consistently left on `.name` as described. Formatter re-wraps look right.

## Two in-scope misses the TS check can't catch

The completeness claim in the description rests on `'name' does not exist` compiler errors. There are two places that use the `'name' in node` narrowing pattern, which compiles cleanly against the v2 `Node` union but silently returns false at runtime for every node:

1. **`packages/visitors-core/src/NodeSelector.ts` ~L36** — `if (name && (!('name' in node) || camelCase(name) !== node.name))`. This means every name-bearing selector (`[programNode]foo`, `[errorNode]${name}`, `[definedTypeNode]bar`, the `selector` argument of `updateAccountsVisitor`/`updateDefinedTypesVisitor`/`updateProgramsVisitor`, `deleteNodesVisitor`, …) will match nothing. Several files in this very PR now build selectors from `.identifier` (`deduplicateIdenticalDefinedTypesVisitor`, `addPdasVisitor`, `updateAccountsVisitor`) and would be no-ops. Needs `'identifier' in node` / `node.identifier`. (Separately, and for a follow-up rather than this PR: `camelCase(name) !== node.identifier` will also need a decision now that v2 identifiers aren't camelCase-normalised — case-fold/underscore-strip compare per the brand docstring, or normalise both sides.)

2. **`packages/visitors-core/src/getDebugStringVisitor.ts` ~L115** — the `default:` branch `return 'name' in node ? [node.name] : [];`. This is the generic fallback that prints the name for `accountNode`, `instructionNode`, `definedTypeNode`, `pdaNode`, `structFieldTypeNode`, enum variants, `variablePdaSeedNode`, `accountValueNode`, `pdaSeedValueNode`, etc. — i.e. most of the debug output. Same fix. (Anchored a nit on the link-node branch a few lines above.)

Worth a quick `rg "'name' in"` across `packages/*/src` to confirm those are the only two.

## Notes for the description / follow-ups

- **`instructionArgumentNode` is also a removed node in v2** — `InstructionNode` now carries `data?: TypeNode` and there's no `InstructionArgumentNode` in `@codama/node-types` (the remaining `packages/nodes/src/InstructionArgumentNode.ts` helper imports a type that no longer exists). This PR still renames `name:` → `identifier:` on `instructionArgumentNode({...})` constructions and `ix.arguments[].name` reads (nodes-from-anchor `InstructionNode.ts`/`InstructionArgumentNode.ts`, `createSubInstructionsFromEnumArgsVisitor`, `setInstructionDiscriminatorsVisitor`, `updateInstructionsVisitor`, `resolve-standalone-pda`, `encode-instruction-arguments`, …). That's harmless — those sites will become struct fields, which do use `identifier` — but it's not in the description's renamed-kinds list and doesn't fit the "removed-node branches are left untouched" rule either. Might be worth a line in the description so the per-package PRs don't treat them as done.
- Other non-rename shape changes I noticed while cross-checking, all correctly out of scope here but easy to lose track of: `fieldDiscriminatorNode.name` → `path` (`dynamic-parsers/discriminators.ts`, `getDebugStringVisitor`, `nodes-from-anchor` `fieldDiscriminatorNode('discriminator')`); `accountFieldValueNode` → `accountDataValueNode` (`resolve-consumed-members.ts`, `resolve-injected-value.ts`).

## For subsequent reviewers

Nothing else needs deep attention — this is the rename it says it is. Once the two `'name' in node` sites are fixed I'm happy to approve.

PR #1167, opened 2026-09-21
[2.x] Adapt errors to Codama v2

lorisleiva:
This PR adapts `@codama/errors` to the v2 node model.

## Changes

`src/context.ts`:
- **`CamelCaseString` → `IdentifierString`** across the ~42 error-context fields that hold node identifiers (v2 identifiers are `IdentifierString`, not casing-branded).
- **Resolver error contexts now span accounts _and_ data.** In v2, instruction arguments are struct fields of `instructionNode.data`, and a data field can carry a contextual default via the inject/provide pattern: `structFieldTypeNode.defaultValue` may be an `injectedValueNode` whose key is fulfilled by `instructionNode.provides` → `providedNode.node` (a contextual value such as `dataValueNode`, `accountValueNode`, `accountBumpValueNode`, …). So a default-value dependency graph — and therefore a cycle — can involve accounts, data fields, or both. `defaultValue` lives on `InstructionAccountNode` and `StructFieldTypeNode` only, so those are the participant types.
  - **`CYCLIC_DEPENDENCY_DETECTED_WHEN_RESOLVING_INSTRUCTION_DEFAULT_VALUES`**: `cycle` retyped `(InstructionAccountNode | InstructionArgumentNode)[]` → `(InstructionAccountNode | StructFieldTypeNode)[]`.
  - **`INVALID_INSTRUCTION_DEFAULT_VALUE_DEPENDENCY`**: `dependency` → `AccountValueNode | DataValueNode`, `dependencyKind` → `'accountValueNode' | 'dataValueNode'` (`dataValueNode` replaces the removed `argumentValueNode`); `dependencyName` → `IdentifierString | PathString` (a `dataValueNode` carries a `path`); `parent` → `InstructionAccountNode | StructFieldTypeNode`, `parentKind` → `'instructionAccountNode' | 'structFieldTypeNode'`, `parentName` → `IdentifierString | PathString`.
- Imports: dropped `ArgumentValueNode`, `InstructionArgumentNode`, `CamelCaseString`; added `DataValueNode`, `IdentifierString`, `PathString`, `StructFieldTypeNode`.

`src/messages.ts`:
- `CYCLIC_...` reworded to drop the participant qualifier (was "accounts and arguments' default values"), since a cycle may run through accounts, data, or both: "resolving the default values of the [$instructionName] instruction".
[...]

PR #1171, opened 2026-09-22
[2.x] Adapt nodes to Codama v2

lorisleiva:
This PR adapts the hand-written helpers and tests of `@codama/nodes` to the v2 node model. The package is fully green: it type-checks, builds, tree-shakes, and 1113/1113 tests pass.

## `src`

Deleted helpers whose v1 concept no longer exists (all with zero surviving consumers, or consumers being rewritten when their own package is adapted):
- `InstructionArgumentNode.ts` — `structTypeNodeFromInstructionArgumentNodes` / `structFieldTypeNodeFromInstructionArgumentNode`. Instruction arguments are now `instructionNode.data` directly.
- `NestedTypeNode.ts` — `resolveNestedTypeNode` / `transformNestedTypeNode` / `isNestedTypeNode` / `assertIsNestedTypeNode`. Wrapper type nodes are replaced by the flat `transforms` array, so resolving is the identity and the guards collapse to `isNode`/`assertIsNode`.
- `NumberTypeNode.ts` — `isSignedInteger` / `isUnsignedInteger` / `isInteger` / `isDecimal`. Subsumed by the `integerTypeNode` / `floatTypeNode` split.
- `getAllInstructionArguments` (in `InstructionNode.ts`) — read the removed `arguments`/`extraArguments`. Its three consumers are rewritten in their own packages to read `instructionNode.data`.

Reworked:
- `EnumTypeNode.isScalarEnum` — the three variant kinds collapsed into `enumVariantTypeNode` with optional `data`; a scalar enum is one whose variants all have no `data`.
- `index.ts` — dropped the exports above and the dead `ENUM_VARIANT_TYPE_NODES` alias (its union no longer exists). The other eight legacy plural aliases stay: `spec-generators` emits them into the generated visitors.

Added:
- `TypeNode.ts` — `addTypeNodeTransforms<T extends TypeNode>(typeNode, transforms): T`. The v2 counterpart of re-wrapping a type in its original wrapper nodes: appends `transforms` after any the node already carries (first is innermost), omits the attribute when empty, returns a frozen copy. Lets a consumer that rebuilds a type node carry the original's transforms across without a double spread.

## `test`

- Deleted the 20 test files for removed nodes/helpers.
- Adapted the 31 surviving test files to v2 (`identifier`, string-encoded numeric values, flat transforms, single `docs` string, `options`-bag ergonomics). Three assertions changed intent because the v1 behaviour is gone: identifier casing is now preserved rather than camelCased (with a companion "rejects invalid identifiers" test); `docs` is a single string; `instructionRemainingAccountsNode` is built from an identifier, not `argumentValueNode`.
- Added tests for every new v2 node (`integerTypeNode`, `floatTypeNode`, `fixedPointTypeNode`, `durationTypeNode`, `enumVariantTypeNode`, `integerValueNode`, `floatValueNode`, `textNode`, `dataValueNode`, `accountDataValueNode`, `sentinelCountNode`, `unitNumberDisplayNode`, the seven transform nodes) plus `addTypeNodeTransforms`. The string-encoded value nodes assert that malformed input throws `CODAMA_ERROR__INVALID_BRANDED_STRING`.

## Follow-ups

- Consumers of the deleted helpers, rewritten as their packages are adapted: `getAllInstructionArguments` (validators, `fillDefaultPdaSeedValuesVisitor`, the visitors-core resolver → read `instructionNode.data` fields); `resolveNestedTypeNode` (identity — use the node); `transformNestedTypeNode` (→ `addTypeNodeTransforms` where a fresh node replaces one that had transforms); `structTypeNodeFromInstructionArgumentNodes` (→ `node.data`).
- `CODAMA_ERROR__UNEXPECTED_NESTED_NODE_KIND` in `@codama/errors` is orphaned now that `assertIsNestedTypeNode` is gone — remove in a later errors touch-up.
- `unwrapTupleEnumWithSingleStructVisitor` and `setNumberWrappersVisitor` target removed kinds wholesale; the visitors PR decides delete-vs-rewrite.


lorisleiva:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/codama-idl/codama/1171?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1171** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/codama-idl/codama/1171?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1162** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1162?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#1158** <a href="https://app.graphite.com/github/pr/codama-idl/codama/1158?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

lorisleiva:
@trevor-cortex

packages/nodes/src/TypeNode.ts
```diff
@@ -0,0 +1,29 @@
...
+ * parameter is not re-derived, as with the generated constructors' own
+ * `as TTransforms` casts.
+ */
+export function addTypeNodeTransforms<T extends TypeNode>(typeNode: T, transforms: readonly TransformNode[]): T {
+    if (transforms.length === 0) return typeNode;
+    const merged = [...(typeNode.transforms ?? []), ...transforms];
+    // Spreading a generic `T` widens to an index-signature object, so the
+    // narrowed node type has to be reasserted through `unknown`.
+    return Object.freeze({ ...typeNode, transforms: merged }) as unknown as T;
+}
```
trevor-cortex, line 34: Nit: the append semantics are right for the documented case (a *fresh* node replacing one that had transforms), but the natural way to "rebuild" a node is often `{ ...old, fields: newFields }` — and that node already carries `old.transforms`, so `addTypeNodeTransforms(rebuilt, old.transforms ?? [])` would duplicate them. Since the visitors migration of `transformNestedTypeNode` is going to reach for this helper a lot, a one-liner in the docblock along the lines of "only pass transforms the node does not already carry; a node derived by spreading the original already has them" would save someone a confusing double-`fixedSizeTransformNode` bug later.

trevor-cortex (approved):
## Summary

Adapts the hand-written side of `@codama/nodes` to the v2 node model. The `src` delta is small and mechanical: four helpers whose v1 concepts no longer exist are deleted (`InstructionArgumentNode.ts`, `NestedTypeNode.ts`, `NumberTypeNode.ts`, `getAllInstructionArguments`), `isScalarEnum` is re-expressed as "no variant carries `data`", the dead `ENUM_VARIANT_TYPE_NODES` alias is dropped, and a single new helper `addTypeNodeTransforms` is added as the flat-transforms replacement for re-wrapping a mapped type. The bulk of the diff is tests: 20 files for removed nodes deleted, 31 adapted (`identifier`, string-encoded numerics, `options` bags, single-string `docs`, flat transforms), and new coverage for every new v2 node plus two new kind-parity typetests.

I read every `src` diff and every test diff, and cross-checked `addTypeNodeTransforms` against the generated `TypeNode`/`StandaloneTypeNode`/`DefinedTypeLinkNode` types — every member of the `TypeNode` union carries `transforms?`, so the generic access is sound, and the "first is innermost" ordering in the docblock matches the generated field docs. The v1 → v2 mapping in the tests is faithful, the three intentional assertion changes are exactly the ones called out in the description, and the surviving `InstructionNode.ts` helpers (`getAllInstructionsWithSubs`, `parseOptionalAccountStrategy`) still line up with the v2 `InstructionNode` shape.

## Things worth a look

- **`addTypeNodeTransforms` double-apply hazard** — the helper is correct for its stated use (carrying transforms onto a *fresh* node), but a consumer that derives the replacement via spread (`{ ...old, fields }`) already has `old.transforms` and would end up with them twice. Left an inline nit suggesting a one-line warning in the docblock, since this is exactly the pattern the `transformNestedTypeNode` migration in visitors will hit.
- **`packages/nodes/README.md` is now stale** — it still uses `numberTypeNode('u64')` / `NumberTypeNode` as its worked example and links to the `v1/docs` spec pages. Not introduced by this PR, but this PR is what makes the example wrong for the package. Fine as a follow-up alongside the other v2 doc touch-ups, or a trivial fix here.
- **Changeset** — removing public exports and adding `addTypeNodeTransforms` is user-facing. If the 2.x line is covered by the seeded major changeset described in `CONTRIBUTING.md` then nothing is needed; otherwise a short changeset listing the removed helpers and the new one would make the eventual CHANGELOG much more useful to people migrating. I couldn't list `.changeset/` from here, so flagging rather than asserting.

- The `CODAMA_ERROR__INVALID_BRANDED_STRING` assertions in the new value-node tests match on the exact human-readable `expected` string (repeated across `ConstantNode`, `AccountDataValueNode`, `DataValueNode`, `EnumVariantTypeNode`, `IntegerValueNode`, `FloatValueNode`). They pass today, but a wording tweak in the brand validators will fan out to six test files. Acceptable coupling for now; just be aware when the brands get touched.
- Several new tests (skip-when-empty arrays, default `endian`/`prefix`/`strategy`, etc.) exercise generated constructors rather than hand-written code. That's consistent with the pre-existing `ProgramNode.test.ts` pattern and cheap documentation of behaviour, so no objection — just don't expect them to catch generator regressions that CI's regenerate-and-diff check wouldn't already.
- The follow-up list in the description is accurate: `CODAMA_ERROR__UNEXPECTED_NESTED_NODE_KIND` is orphaned as of this PR, and `unwrapTupleEnumWithSingleStructVisitor` / `setNumberWrappersVisitor` target kinds that no longer exist.

